PR #6, opened 2020-01-08
Correct Windows MSYS2 paths

MSP-Greg:
Add `bin` to Windows MSYS2 paths.  I just edited the files for now...

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -6521,7 +6521,7 @@ function setupPath(msys2, rubyPrefix) {
   path = path.filter(e => !e.match(/\bRuby\b/))
 
   // Add MSYS2 in PATH
-  path.unshift(`${msys2}\\mingw64`, `${msys2}\\usr`)
+  path.unshift(`${msys2}\\mingw64\\bin`, `${msys2}\\usr\\bin`)
 
   // Add the downloaded Ruby in PATH
   path.unshift(`${rubyPrefix}\\bin`)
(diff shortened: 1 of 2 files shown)


eregon:
Oops, good catch, thanks for the fix!

eregon:
FWIW, the `dist` file is generated by `yarn run package`

MSP-Greg:
JFYI, code that's part of RubyInstaller2 ('RI2') requires 'devkit', which sets the path for the MSYS2 bin folders.  This is done automatically when installing an extension gem.

But, many repos don't load devkit in their build scripts for CI.

I may have a look at adding it via mkmf.rb, and there's also some mswin issues...

eregon:
> JFYI, code that's part of RubyInstaller2 ('RI2') requires 'devkit', which sets the path for the MSYS2 bin folders. This is done automatically when installing an extension gem.

Right, that's why it worked when trying to install `json` (and that failed without MSYS2, I checked).

> But, many repos don't load devkit in their build scripts for CI.

Seems fine to me for now to add those directories to PATH.

MSP-Greg:
I'm sorry, I was trying to explain my point of view for placing OS specific 'adjustments' in the code here.

Two main issues, the first is packages, and how to install them, especially problems like OpenSSL.  As mentioned, I released msys2-actions for that.

The second is the path problem.  RubyInstaller2 hooks into RubyGems with `rubygems/defaults/operating_system.rb`, but that doesn't help when using mkmf.rb in CI.  There's also some issues with mswin builds. I may make some changes and do a PR in ruby/ruby.

Thanks for all your work on this.  It will be interesting to see how it moves forward with GitHub, along with the Windows virtual environment MSYS2 issue...

PR #8, opened 2020-01-09
Set CLASSPATH to empty for JRuby

MSP-Greg:
This sets CLASSPATH to an empty string before the download/installation.

Not sure whether this is considered a valid thing to do, but the 'Ruby Info' step in https://github.com/MSP-Greg/use-ruby-action-info/runs/381796430 completes with it.

Using yarn now.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,9 @@ async function run() {
     if (platform === 'windows-latest') {
       rubyPrefix = await windows.downloadExtractAndSetPATH(ruby)
     } else {
+      if (ruby.startsWith('jruby')) {
+        core.exportVariable('CLASSPATH', '')
+      }
       rubyPrefix = await downloadAndExtract(platform, ruby)
       core.addPath(`${rubyPrefix}/bin`)
     }
--- a/dist/index.js
+++ b/dist/index.js
@@ -1399,6 +1399,9 @@ async function run() {
     if (platform === 'windows-latest') {
       rubyPrefix = await windows.downloadExtractAndSetPATH(ruby)
     } else {
+      if (ruby.startsWith('jruby')) {
+        core.exportVariable('CLASSPATH', '')
+      }
       rubyPrefix = await downloadAndExtract(platform, ruby)
       core.addPath(`${rubyPrefix}/bin`)
     }


eregon:
I think that should be solved in https://github.com/actions/virtual-environments/issues/242
I'm thinking such a workaround here would be too much of a hack.
But maybe we have no choice if that takes too long to resolve.

Users can workaround with
```yaml
  env:
    CLASSPATH: ""
```
until the issue above is fixed.

index.js
```diff
@@ -13,11 +13,15 @@ async function run() {
   try {
     const platform = getVirtualEnvironmentName()
     const ruby = await getRubyEngineAndVersion(core.getInput('ruby-version'))
+    const reJRuby = /^jruby/i
 
     let rubyPrefix
     if (platform === 'windows-latest') {
       rubyPrefix = await windows.downloadExtractAndSetPATH(ruby)
     } else {
+      if (reJRuby.test(ruby)) {
```
eregon: Could you change it to `if (ruby.startsWith('jruby')) {` ?
That's good enough, `getRubyEngineAndVersion` ensures the engine name is lower case.

eregon:
@MSP-Greg  Changing my mind, let's merge this.

https://github.com/actions/virtual-environments/issues/242 has been opened for 21 hours and no response from GitHub so far it seems.

index.js
MSP-Greg: Done

eregon:
Thanks, it works as seen in https://github.com/MSP-Greg/use-ruby-action/runs/381846122

Let's merge this for now to avoid more builds failings.
Might as well workaround closer to the source to avoid breakage.
I hope that GitHub will resolve this soon too.

MSP-Greg:
Seen this type of thing before on AppVeyor & Travis.  If it breaks once, it might break again.  Since I'm java challenged, I'll defer to you.

JFYI, I checked OpenSSL dll's on Windows earlier.  There are four copies that appear in the standard path.  Granted, a repo's CI should account for that, but if one is testing locally with no issues, it's a easy thing to overlook.

MSP-Greg:
\<off-topic\>

Re the above paragraph, the same also happens with multiple copies of build tools in the path.

This and the problem here are things that people should be aware of, but everyone doesn't know JRuby well, nor Windows.  Hence, I tend to fix it in the action, and save the support time...

Re Puma, my goal was actually to check that it built with the same OpenSSL version as the Ruby it was built with, specifically Ruby 2.3 & OpenSSL 1.0.2.  With Windows builds, the MSYS2 OpenSSL package has to match the Ruby package used for the built.

That is the main reason I have code in `actions-ruby` and `msys2-action` to handle OpenSSL, as 2.4 uses 1.0.2.  Note that I updated `msys2-action` to work with this, Puma PR uses it to update the gcc tools, add ragel, and install the correct OpenSSL.

I was pleasantly surprised that `ruby-build` includes all the OpenSSL files needed to build.  Hence, Puma built using Ruby 2.3 was using 1.0.2 for build & runtime...

PR #14, opened 2020-01-30
Use SSD drive for Windows Ruby location

MSP-Greg:
From a bit of quick research, the systems used for Actions have both a HD and an SSD.  Added code to place the Windows Ruby on the SSD, assuming it is `d:`

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -25,9 +25,11 @@ export async function install(platform, ruby) {
   }
   const base = url.slice(url.lastIndexOf('/') + 1, url.length - '.7z'.length)
 
+  const drv = (process.env['GITHUB_WORKSPACE'] || 'C')[0]
+
   const downloadPath = await tc.downloadTool(url)
-  await exec.exec(`7z x ${downloadPath} -xr!${base}\\share\\doc -oC:\\`)
-  const rubyPrefix = `C:\\${base}`
+  await exec.exec(`7z x ${downloadPath} -xr!${base}\\share\\doc -o${drv}:\\`)
+  const rubyPrefix = `${drv}:\\${base}`
 
   const [hostedRuby, msys2] = await linkMSYS2()
   const newPath = setupPath(msys2, rubyPrefix)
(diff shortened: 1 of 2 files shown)


eregon:
Could you give some points to that research?

windows.js
```diff
@@ -25,9 +25,11 @@ export async function install(platform, ruby) {
   }
   const base = url.slice(url.lastIndexOf('/') + 1, url.length - '.7z'.length)
 
+  const drv = (process.env['RUNNER_WORKSPACE'] || 'C')[0]
```
eregon: Should we use `GITHUB_WORKSPACE` instead, which an officially documented env var?
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables

MSP-Greg:
Info on VM:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#cloud-hosts-for-github-hosted-runners

States:

Each virtual machine has the same hardware resources available.

* 2-core CPU
* 7 GB of RAM memory
* 14 GB of SSD disk space

Uses Standard_DS2_v2:
https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes-general#dsv2-series

On a Windows VM:
```
PS C:\> Get-CimInstance Win32_LogicalDisk

DeviceID DriveType ProviderName VolumeName        Size         FreeSpace
-------- --------- ------------ ----------        ----         ---------
A:       2                                                     
C:       3                      Windows           136912564224 8117616640
D:       3                      Temporary Storage 15031267328  13100244992
```

The numbers line up with D: being the SSD.

windows.js
MSP-Greg: Ok.  Done.

eregon:
Great, thank you for the PR, let's merge it :)

MSP-Greg:
JFYI,

The whole step of downloading and extracting didn't seem to change much time-wise, but SSD's have got to be faster than HD's...

Working on fully tested master binaries, I'll probably use an embedded action for it.  Maybe when I'm done we can think about non MRI Rubies...

slonopotamus:
It is forbidden to use magic outside of Hogwarts!

Issue #19, opened 2020-02-09
setup-ruby removes chocolatey from PATH


slonopotamus:
See trivial workflow that reproduces the problem: https://github.com/slonopotamus/asciidoctor-diagram/runs/434520900?check_suite_focus=true

Before setup-ruby actions is called, there's `C:\ProgramData\Chocolatey\bin` on PATH. But it is gone after setup-ruby is called.

For the reference, here's workflow file:

```
name: CI
on: [push, pull_request]
jobs:
  test:
    runs-on: windows-latest
    steps:
      - run: |
          echo "Before setup-ruby"
          $Env:Path
          choco --version
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: 2.7
      - run: |
          echo "After setup-ruby"
          $Env:Path
          choco --version
```

MSP-Greg:
@slonopotamus 

I'm essentially responsible for the code that does that.

On CI systems, there's often issues with both multiple copies of applications being available and also 'dll resolution hell'.  It's one thing to have an error showing that a dll can't be found, it's much worse if a different dll is loaded and then has unexpected results.

Then again, one should be aware of issues like this, but a lot of macOS or *nix users won't expect them.

Regardless, I don't have an issue with removing the code.  Can your workflows work around the path changes?

JFYI, most publicly facing Ruby MinGW builds use a manifest for dll resolution, and are bundled with their own dlls.  But, the ruby/ruby CI is done without one.

eregon:
Thanks for the report.
@MSP-Greg So what do you think, should we just remove Chocolatey from that line?
https://github.com/ruby/setup-ruby/blob/30011b6444339b8df773cfc724e1631d805a5e64/windows.js#L68

We could also have an option to not filter the PATH at all (on Windows).
But for example having another `ruby` in PATH leads to very confusing errors, so at the very least we should remove that one.

mingw64 should likely also be removed as it would conflict with msys2.
Strawberry Perl is also a sort of MinGW if I understand correctly, so is similar.

Not sure about CMake and OpenSSL. Maybe msys2 already provide these 2?

slonopotamus:
> Can your workflows work around the path changes?

Sure, I can workaround this by reordering steps. It was just really unexpected and not very trivial to diagnose.

eregon:
> It was just really unexpected and not very trivial to diagnose.

I'm sorry about that.

[...]

PR #20, opened 2020-02-09
Fix PATH on Windows

eregon:
* See https://github.com/ruby/setup-ruby/issues/19

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -62,7 +62,8 @@ async function linkMSYS2() {
 }
 
 export function setupPath(msys2, rubyPrefix) {
-  let path = process.env['PATH'].split(';')
+  const originalPath = process.env['PATH'].split(';')
+  let path = originalPath.slice()
 
   // Remove conflicting dev tools from PATH
   path = path.filter(e => !e.match(/\b(Chocolatey|CMake|mingw64|OpenSSL|Strawberry)\b/))
@@ -77,6 +78,13 @@ export function setupPath(msys2, rubyPrefix) {
 
   // Add the downloaded Ruby in PATH
   path.unshift(`${rubyPrefix}\\bin`)
+
+  console.log("Entries removed from PATH to avoid conflicts with MSYS2 and Ruby:")
+  for (const entry of originalPath) {
+    if (!path.includes(entry)) {
+      console.log(entry)
+    }
+  }
 
   const newPath = path.join(';')
   core.exportVariable('PATH', newPath)
--- a/windows.js
+++ b/windows.js
@@ -29,7 +29,7 @@ export async function install(platform, ruby) {
   const drive = (process.env['GITHUB_WORKSPACE'] || 'C')[0]
 
   const downloadPath = await tc.downloadTool(url)
-  await exec.exec(`7z x ${downloadPath} -xr!${base}\\share\\doc -o${drive}:\\`)
+  await exec.exec('7z', ['x', downloadPath, `-xr!${base}\\share\\doc`, `-o${drive}:\\`], { silent: true })
   const rubyPrefix = `${drive}:\\${base}`
 
   const [hostedRuby, msys2] = await linkMSYS2()
--- a/windows.js
+++ b/windows.js
@@ -65,9 +65,6 @@ export function setupPath(msys2, rubyPrefix) {
   const originalPath = process.env['PATH'].split(';')
   let path = originalPath.slice()
 
-  // Remove conflicting dev tools from PATH
-  path = path.filter(e => !e.match(/\b(Chocolatey|CMake|mingw64|OpenSSL|Strawberry)\b/))
-
   // Remove default Ruby in PATH
   path = path.filter(e => !e.match(/\bRuby\b/))
 
@@ -79,7 +76,7 @@ export function setupPath(msys2, rubyPrefix) {
   // Add the downloaded Ruby in PATH
   path.unshift(`${rubyPrefix}\\bin`)
 
-  console.log("Entries removed from PATH to avoid conflicts with MSYS2 and Ruby:")
+  console.log("Entries removed from PATH to avoid conflicts with Ruby:")
   for (const entry of originalPath) {
     if (!path.includes(entry)) {
       console.log(entry)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,7 @@ name: Test this action
 on:
   push:
     branches-ignore:
+    - master
     - v1
     tags-ignore:
     - '*'
--- a/dist/index.js
+++ b/dist/index.js
@@ -7929,7 +7929,8 @@ async function linkMSYS2() {
 }
 
 function setupPath(msys2, rubyPrefix) {
-  let path = process.env['PATH'].split(';')
+  const originalPath = process.env['PATH'].split(';')
+  let path = originalPath.slice()
 
   // Remove conflicting dev tools from PATH
   path = path.filter(e => !e.match(/\b(Chocolatey|CMake|mingw64|OpenSSL|Strawberry)\b/))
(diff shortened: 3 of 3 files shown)


windows.js
```diff
@@ -62,7 +62,8 @@ async function linkMSYS2() {
 }
 
 export function setupPath(msys2, rubyPrefix) {
-  let path = process.env['PATH'].split(';')
+  const originalPath = process.env['PATH'].split(';')
+  let path = originalPath.slice()
 
   // Remove conflicting dev tools from PATH
   path = path.filter(e => !e.match(/\b(Chocolatey|CMake|mingw64|OpenSSL|Strawberry)\b/))
```
eregon: @MSP-Greg @slonopotamus What do you think is the best way to proceed?
* Remove this line entirely
* Make it an option to clean to path, and choose a default value
* Remove only `Chocolatey` from this line

What are the risks if we keep any of them in PATH?
Ideally I think a `setup-*` action shouldn't remove anything in PATH except a conflicting tool (Ruby, done below).

MSP-Greg: > What are the risks if we keep any of them in PATH?

Confused people.

> Ideally I think a setup-* action shouldn't remove anything in PATH except a conflicting tool.

I think that would be best.  If people want to use MSYS2, they can use a windows specific action or MSYS2 commands.

But as mentioned, I don't know if a windows specific option to clean the path and enable MSYS2 might be helpful.   It's easy, and that's what the majority of Windows Ruby CI (that requires MinGW compile tools) is doing.

eregon: > But as mentioned, A windows specific option to clean the path and enable MSYS2 might be helpful. It's easy, and that's what the majority of Windows Ruby CI (that requires MinGW compile tools) is doing.

I think most people expect MSYS2 to be available by default.

And I would think there are cases where one wants both MSYS2 and installing packages with `choco`? Or is that always problematic/a mistake?

With that thinking, if we add an option it would just be an option to clean the PATH, and otherwise leave it untouched (except adding/removing Ruby). I'd still add MSYS2 no matter what, as I don't think there is a need for "no MSYS2" (and therefore "cannot compile C extensions").

eregon: @slonopotamus For context, what do you want to install with Chocolatey?
Is it something that could be installed with MSYS2's pacman instead?

Do you need to be able to install C extensions on Ruby on Windows?

@MSP-Greg Would installing via pacman generally work better than with Chocolatey for use with Ruby?

MSP-Greg: @eregon 

> Would installing via pacman generally work better than with Chocolatey for use with Ruby?

I've helped a lot of extension gems/repos with Windows CI, and I don't recall ever using a 'Chocolatey package' for compiling.  But, some repos may use it for other requirements.

slonopotamus:
> For context, what do you want to install with Chocolatey?
Is it something that could be installed with MSYS2's pacman instead?
>
> Do you need to be able to install C extensions on Ruby on Windows?

Just some random software, totally Ruby-unrelated: graphviz, gnuplot, imagemagick.

What I am actually trying to do is to convert [asciidoctor-diagram appveyor.yml](https://github.com/asciidoctor/asciidoctor-diagram/blob/master/appveyor.yml) to GH-Actions.

MSP-Greg:
@slonopotamus

> What I am actually trying to do is to convert asciidoctor-diagram appveyor.yml to GH-Actions.

32 bit Ruby?  Do you need that?

I've asked for 32 bit MSYS2 tools on Actions, but they can't seem to decide about MSYS2.

slonopotamus:
> 32 bit Ruby?

Err... I don't understand where you see that. As I said, I use Chocolatey for non-Ruby software.

MSP-Greg:
@slonopotamus

The matrix (shown below) is all 32 bit Ruby builds.  Normally, 64 bit has a `-x64` suffix.  For instance, see https://github.com/ruby/psych/blob/master/appveyor.yml.

If you're not compiling anything for Ruby, it doesn't matter, but Actions is only 64 bit Rubies.

```
environment:
  matrix:
    - ruby_version: "25"
    - ruby_version: "24"
    - ruby_version: "23"
```

eregon:
> Just some random software, totally Ruby-unrelated: graphviz, gnuplot, imagemagick.

There are MSYS2 packages for all these 3: https://packages.msys2.org/base

I think the advantage of using those (anyone correct me if I'm wrong, I'm not very familiar with Windows) is if any Ruby gem compiles against one of them there would be a much higher chance for it to work, since it would use the same kind of toolchain Ruby was compiled with.

What does Chocolatey uses for the build toolchain? mingw-x64? MSYS2? Something else?

slonopotamus:
Okay, no, I'm not trying to use 32-bit Ruby. Here's my work-in-progress workflow file (it doesn't work properly yet): https://github.com/slonopotamus/asciidoctor-diagram/blob/gh-actions/.github/workflows/ci.yml

MSP-Greg:
>What does Chocolatey uses for the build toolchain?

That's not specific to Chocolatey.  One can install MSYS2 with Chocolatey.  So, what it has is what GH decided to pre-install (I think).

slonopotamus:
> What does Chocolatey uses for the build toolchain?

Chocolatey is used to just install Windows software. It's not a build environment.

> There are MSYS2 packages for all these 3
The point about Chocolatey is that it is a convenient way to install and update programs on Windows without the need to go to multiple websites searching for installers. You may see Chocolatey on non-programmer machines, unlike MSYS2 that is kinda specific thing. You would normally use MSYS2 tools from within a MSYS2 shell. This is not the case of Chocolatey, it mostly provides native Windows software.

eregon:
Right, but those precompiled Chocolatey packages must have been compiled with some toolchain.
I guess C toolchains might be mostly compatible, but I wouldn't have much hope for C++.

As seen in https://github.com/eregon/setup-ruby-test/runs/434763137, there is a toolchain pre-installed in Chocolatey. I'm not sure which one it is but I'd guess MinGW.

OK, I agree we should let users use Chocolatey if they want to.

So how about not removing Chocolatey from the PATH, and see how that goes?
If there are weird errors due to conflicts with MSYS2 we'll need to reconsider this and e.g., have an option about this.

I can also add an option to remove nothing but Ruby from the PATH.
That seems a good safeguard.

@MSP-Greg 
>> What are the risks if we keep any of them in PATH?

> Confused people.

Could you give an example of what could go wrong if we remove nothing but Ruby from the PATH, and still have MSYS2 paths at the start of the PATH?

slonopotamus:
> Right, but those precompiled Chocolatey packages must have been compiled with some toolchain.

Chocolatey doesn't provide build infrastructure, it only handles installation step. You can just wrap pre-existing MSI package in Chocolatey and you don't need to know how and with what tools that MSI was created.

> So how about not removing Chocolatey from the PATH, and see how that goes?

I think everything should be fine as long as you add whatever is needed for Ruby at the beginning of PATH.

eregon:
OK, I propose to just not remove anything from PATH then, except the system Ruby (no point to have 2 Ruby in PATH).
That's what I tried in c30207dd64f0aead4c021750963b2439388e99c1 and the (limited) tests here seem fine. At least installing `json` keeps working.

@MSP-Greg Is that OK with you?
Do you think we need/should have an option to clean the PATH as previous behavior?
I think I would still want that option disabled by default, by the reasoning that a `setup-*` action should not remove tools, only add (or replace for the specific tool it's adding, here Ruby).

MSP-Greg:
> Is that OK with you?

Yes

> Do you think we need/should have an option to clean the PATH as previous behavior?

I'd suggest waiting on it.  We can always add to the README about windows compile issues...

eregon:
Merged and added notes in the README about Windows in 4f28d1c82e3c9bf7dd6eb6fb060c3066b2125349

PR #23, opened 2020-02-17
Use runner.os for testing against one platform

JuanitoFatas:
Hi 👋 

The `${{ matrix.os }}` is empty when only testing on one platform. We can suggest to use `${{ runner.os }}` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ You can cache the installed gems with these two steps:
         bundle install --jobs 4 --retry 3
 ```
 
+When using a single OS, replace `${{ matrix.os }}` with the OS.  
 When using a single job with a Ruby version, replace `${{ matrix.ruby }}` with the Ruby version.  
 When using `.ruby-version`, replace `${{ matrix.ruby }}` with `${{ hashFiles('.ruby-version') }}`.  
 When using `.tool-versions`, replace `${{ matrix.ruby }}` with `${{ hashFiles('.tool-versions') }}`.


eregon:
Hello,

`runner.os` is also used in the original documentation:
https://github.com/actions/cache/blob/master/examples.md#ruby---bundler

I'm unsure if it's correct though, because that returns `Linux` for both Ubuntu 16.04 and Ubuntu 18.04, according to the docs: https://help.github.com/en/actions/reference/contexts-and-expression-syntax-for-github-actions#runner-context

We can indeed see it's `Linux` in https://github.com/ruby/ruby/runs/449747370:
```
Dump runner context
{
  "os": "Linux",
  "tool_cache": "/opt/hostedtoolcache",
  "temp": "/home/runner/work/_temp",
  "workspace": "/home/runner/work/ruby"
}
```

I can easily imagine some libraries have different versions between these two Ubuntu versions, and so sharing the cache for both seems incorrect.

For the single OS case, when switching between Ubuntu 16.04 and Ubuntu 18.04 I think it could break if using `runner.os`.

README.md
```diff
@@ -120,8 +120,9 @@ You can cache the installed gems with these two steps:
         bundle install --jobs 4 --retry 3
 ```
 
-When using a single job with a Ruby version, replace `${{ matrix.ruby }}` with the Ruby version.  
-When using `.ruby-version`, replace `${{ matrix.ruby }}` with `${{ hashFiles('.ruby-version') }}`.  
+When using a single os, replace `${{ matrix.os }}` with `${{ runner.os }}`.
+When using a single job with a Ruby version, replace `${{ matrix.ruby }}` with the Ruby version.
```
eregon: Could you keep the trailing spaces here and below? They force a newline in rendered Markdown.

README.md
```diff
@@ -120,8 +120,9 @@ You can cache the installed gems with these two steps:
         bundle install --jobs 4 --retry 3
 ```
 
-When using a single job with a Ruby version, replace `${{ matrix.ruby }}` with the Ruby version.  
-When using `.ruby-version`, replace `${{ matrix.ruby }}` with `${{ hashFiles('.ruby-version') }}`.  
+When using a single os, replace `${{ matrix.os }}` with `${{ runner.os }}`.
```
eregon:
```suggestion
When using a single OS, replace `${{ matrix.os }}` with the OS.  
```

JuanitoFatas:
Thanks for your detailed explanations. I‘ve updated per your suggestion.

Issue #22, opened 2020-02-17
Windows - ruby-head, ruby-mswin


MSP-Greg:
@eregon

I got the mingw work done in a [branch](https://github.com/MSP-Greg/ruby-setup-ruby/commits/windows-ruby-head), it was nice to run yarn, yarn run lint, & yarn run package and see only one line changed in dist/index.js.  I did update yarn.lock, if that's ok.

Question - should the mswin ruby-version string be 'ruby-mswin'?  Obviously, if one is using a single workflow/job layout in Actions, one would need to exclude all OS's except Windows.

It's easy to add, as it's a hard URI, but I'd like to think about whether MSYS2 should be activated or not.  I'm thinking not.  I'm also wondering whether the 'package action' should be available with it's addition/release.  Remember, the main reason I created the build was for stdlib extension gems that are in the Ruby account/organization.

**Off topic Windows stuff:**

Please see:
https://github.com/MSP-Greg/ruby-setup-ruby/runs/449427016?check_suite_focus=true#step:5:8

I added that line, it shows the gcc version that the Ruby build was compiled with.  Note:
```
(Rev2, Built by MSYS2 project) 9.2.0
```
Now, in the same step, note line 19:
```
(Rev2, Built by MSYS2 project) 8.3.0
```

If you check the most recent action at [MSP-Greg/ruby-setup-ruby-info](https://github.com/MSP-Greg/ruby-setup-ruby-info/actions), you'll see that all the recent Ruby binaries were compiled with gcc 9.2.0 Rev 2 (see Ruby info step, ~ line 10 gcc info:).  But, the DevKit version, which is shown by 'ridk version' is 8.2.0 Rev 2.

That is the reason I've always had code to update the gcc tools in my Windows actions.  But, importantly, updating gcc tools can cause issues with older Ruby binaries, like 2.5.0 or 2.6.0.  Long story, it's due adding libssp (Smashing Stack Protector) to gcc in one of the 9.2.0 releases.

Finally, note line 15:
```
package_version: ruby-loco RI2 2.7.0-1
```
I modified a file to add the 'ruby-loco', and that is the easiest way to determine where the build came from.

**Off topic Actions:**

In ruby-loco I have an embedded action that handles the release asset uploading.  I prefer to have hard URI, so I set it up as follows:

1. Upload replacement build asset with filename 'new_ruby_mingw.7z'
2. Delete current asset
3. Rename 'new_ruby_mingw.7z' to 'ruby_mingw.7z'

[...]

Issue #21, opened 2020-02-13
Integrate system operations into setup-ruby


MSP-Greg:
Thinking about what 's needed for some of the repos I've worked in, I'm wondering if integrating some of the MSYS2 related things I've done in msys2-action and actions-ruby could be integrated here.

For instance, inputs could be added like:
```
with:

  pkg: ragel
  macOS:
  mingw:
  ubuntu:
```

**EDIT**: for instance, all the code needed for ragel
```
      - name: ubuntu & macos - install ragel
        if:   startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
        run:  |
          if [ "${{ matrix.os }}" == "macos" ]; then
            brew install ragel
          else
            sudo apt-get -qy install ragel
          fi
```
'ragel' is a package with a cross platform name, so code could install it based on the platform used.  For packages without common names, the OS specific names could be used in the respective inputs.

I recall somewhere seeing something like 'brew update all' or a similar command, which for Windows/MSYS2 might be the equivalent of updating the gcc tools.  Just some general thoughts...

eregon:
I've been thinking about this too, it's quite verbose in GitHub Actions to install a package across platforms, even if it has the same name.

I'm not sure this fits in this action though, wouldn't it fit better in some `install-packages` (or `setup-packages` maybe) action?
```
- uses: unknown/install-packages
  with:
    packages: ragel openssl
```

A [quick search](https://github.com/marketplace?utf8=%E2%9C%93&type=actions&query=install+package) shows https://github.com/marketplace/actions/get-package which does something a bit like that.

eregon:
I think `pkg` is the name of some package manager, so I think it's better to use the full name.

MSP-Greg:
> I'm not sure this fits in this action though

1. Using two different actions means two steps, two repos to download, two fully packaged js files, each with very similar dependency code.

2. If no packages are needed, it's no different than what's used now.

Maybe look at it as the one action needed to configure Ruby and the system for the CI task?

Also, I've had a few workflows with a lot of steps, so the idea of keeping them to a minimum is good.

It's not something I fell strongly about.  Really, all people care about is the equivalent of `rake test` or any failing previous steps...

eregon:
I think the overhead of using a separate action is very small in practice, probably less than 1 second.

[...]

PR #24, opened 2020-02-18
Add ruby-loco mingw and mswin

MSP-Greg:
1. Add ruby-loco mingw and mswin
2. Adjust workflow for mswin - 'ridk version' & 'c extension' steps
3. Allow head, mingw, & mswin to be used without ruby prefix
4. win_names.yml_ tests the above, fails on some non Windows jobs deliberately

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,24 +9,18 @@ on:
     paths-ignore:
     - README.md
 jobs:
-  test:
+  common:
     strategy:
       fail-fast: false
       matrix:
         os: [ ubuntu-16.04, ubuntu-18.04, macos-latest, windows-latest ]
         # Use various version syntaxes here for testing
-        ruby: [ .ruby-version, .tool-versions, 2.2, 2.3, ruby-head, jruby, jruby-head, truffleruby, truffleruby-head, rubinius ]
+        ruby: [ .ruby-version, .tool-versions, 2.2, 2.3, ruby-head, jruby, jruby-head, truffleruby, truffleruby-head ]
         exclude:
         - os: windows-latest
           ruby: truffleruby
         - os: windows-latest
           ruby: truffleruby-head
-        - os: ubuntu-16.04
-          ruby: rubinius
-        - os: macos-latest
-          ruby: rubinius
-        - os: windows-latest
-          ruby: rubinius
     name: ${{ matrix.os }} ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
     steps:
@@ -34,9 +28,20 @@ jobs:
     - uses: ./
       with:
         ruby-version: ${{ matrix.ruby }}
-    - run: ruby --version
-    - run: ridk version
-      if: matrix.os == 'windows-latest' && !startsWith(matrix.ruby, 'jruby') && !startsWith(matrix.ruby, '2.3') && !startsWith(matrix.ruby, '2.2')
+    - name: ruby version and gcc
+      run: |
+        ruby --version
+        ruby -e "puts 'build gcc: ' + RbConfig::CONFIG.fetch('CC_VERSION_MESSAGE', 'unknown')[/[^\n]+/]"      
+    - name: ridk version (mingw)
+      if: matrix.os == 'windows-latest'
+      run: |
+        # run 'ridk version'
+        $abi, $plat = $(ruby -e "STDOUT.write RbConfig::CONFIG['ruby_version'] + ' ' + RUBY_PLATFORM").split(' ')
+        if (($abi -ge '2.4') -and $plat.Contains('mingw')) {
+          ridk version
+        } else {
+          echo 'ridk is unavailable'
+        }
     - name: Subprocess test
       run: ruby test_subprocess.rb
     - name: OpenSSL version
@@ -45,6 +50,62 @@ jobs:
       run: ruby -ropen-uri -e 'puts open(%{https://rubygems.org/}) { |f| f.read(1024) }'
     - name: C extension test
       run: gem install json:2.2.0 --no-document
+    - run: bundle --version
+    - run: bundle install
+    - run: bundle exec rake --version
+
+  specials:
+    strategy:
+      fail-fast: false
+      matrix:
+        cfg:
+          - { ruby: rubinius, os: ubuntu-18.04   }
+          - { ruby: mingw   , os: windows-latest }
+          - { ruby: mswin   , os: windows-latest }
+    name: ${{ matrix.cfg.os }} ${{ matrix.cfg.ruby }}
+    runs-on: ${{ matrix.cfg.os }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./
+      with:
+        ruby-version: ${{ matrix.cfg.ruby }}
+    - name: ruby version and gcc
+      run: |
+        ruby --version
+        ruby -e "puts 'build gcc: ' + RbConfig::CONFIG.fetch('CC_VERSION_MESSAGE', 'unknown')[/[^\n]+/]"      
+    - name: ridk version (mingw)
+      if: matrix.cfg.os == 'windows-latest'
+      run: |
+        # run 'ridk version'
+        $abi, $plat = $(ruby -e "STDOUT.write RbConfig::CONFIG['ruby_version'] + ' ' + RUBY_PLATFORM").split(' ')
+        if (($abi -ge '2.4') -and $plat.Contains('mingw')) {
+          ridk version
+        } else {
+          echo 'ridk is unavailable'
+        }
+    - name: Subprocess test
+      run: ruby test_subprocess.rb
+    - name: OpenSSL version
+      run: ruby -ropenssl -e 'puts OpenSSL::OPENSSL_LIBRARY_VERSION'
+    - name: OpenSSL test
+      run: ruby -ropen-uri -e 'puts open(%{https://rubygems.org/}) { |f| f.read(1024) }'
+    - name: C extension test
+      if: endsWith(matrix.cfg.ruby, 'mswin') == false
+      run: gem install json:2.2.0 --no-document
+    - name: C extension test (mswin cmd)
+      if: endsWith(matrix.cfg.ruby, 'mswin')
+      shell: cmd
+      run: |
+        call %VCVARS%
+        gem install json:2.2.0 --no-document
+    - name: C extension test (mswin ps1)
+      if: endsWith(matrix.cfg.ruby, 'mswin')
+      run: |
+        cmd.exe /c "$env:VCVARS && set" | Foreach-Object {
+          $p, $v = $_.split('=')
+          Set-Item -path env:$p -value $v
+        }
+        gem install json:2.2.0 --no-document
     - run: bundle --version
     - run: bundle install
     - run: bundle exec rake --version
--- a/index.js
+++ b/index.js
@@ -46,7 +46,7 @@ function parseRubyEngineAndVersion(rubyVersion) {
   }
 
   let engine, version
-  if (rubyVersion.match(/^\d+/)) { // X.Y.Z => ruby-X.Y.Z
+  if (rubyVersion.match(/^(\d+|head|mingw|mswin)/)) { // X.Y.Z => ruby-X.Y.Z
     engine = 'ruby'
     version = rubyVersion
   } else if (!rubyVersion.includes('-')) { // myruby -> myruby-stableVersion
@@ -55,7 +55,6 @@ function parseRubyEngineAndVersion(rubyVersion) {
   } else { // engine-X.Y.Z
     [engine, version] = rubyVersion.split('-', 2)
   }
-
   return [engine, version]
 }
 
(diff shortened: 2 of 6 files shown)


eregon:
Thank you, I'll review this tomorrow.

MSP-Greg:
No hurry.  I'm thinking about the package action...

ioquatix:
Here is how I did packages in rack.

https://github.com/rack/rack/blob/2b22be07f189fd852fb66a601573c38439b1f7b4/.github/workflows/development.yml#L27-L33

It's not that great but it works.

I think I prefer something simpler where possible, e.g.

```
packages:
  - foo
  - bar
  - baz
```

MSP-Greg:
I thought a generic 'packages' might be helpful, but I doubt many packages have a universal name.

Using your rack example, maybe something like:
```
- name: install packages
  uses: ruby/setup-ruby-packages@v1
  with:
    sudo: libfcgi-dev libmemcached-dev
    brew: fcgi libmemcached
```

Normal Windows Ruby (mingw) has MSYS2 packages and MinGW packages, so for example, a generic package input wouldn't detail whether openssl was the openssl MSYS2 package, or the mingw-w64-x86_64-openssl MinGW package...

MSP-Greg:
A little off topic, but...

One of the issues here and with custom actions is determining what the environment is, whether it be OS, image version, Ruby version/platform, etc.

See https://github.com/actions/virtual-environments/pull/407#issue-375349162

An ENV variable `ImageOS` will be added to the next image releases that will contain the Ubuntu or Windows image type.  There's a table in the comment.  macOS looks like it will need to checked in a different manner.  I mentioned this because I forget stuff...

eregon:
It would be best to keep the packages discussion in one place: https://github.com/ruby/setup-ruby/issues/21
Maybe I can reopen that issue if it helps, although I don't plan to implement that in this action?

Sorry, I got too busy today, will review it when I get some time.

.github/workflows/test.yml
```diff
@@ -35,16 +45,32 @@ jobs:
...
     - name: OpenSSL test
       run: ruby -ropen-uri -e 'puts open(%{https://rubygems.org/}) { |f| f.read(1024) }'
     - name: C extension test
+      if: endsWith(matrix.ruby, 'mswin') == false
       run: gem install json:2.2.0 --no-document
+    - name: C extension test (mswin)
+      shell: cmd
+      if: endsWith(matrix.ruby, 'mswin')
+      run: |
+        call "C:\Program Files (x86)\Microsoft Visual Studio\${{ matrix.vs }}\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
eregon: It seems quite inconvenient to have to use this long line before installing any C extension.
Is this something the action could setup once and for all? Or it has to be done in each shell session?

I think very few Rubyists expect having to anything extra before being able to install C extensions.

index.js
```diff
@@ -7,6 +7,10 @@ async function run() {
     const platform = getVirtualEnvironmentName()
     const [engine, version] = parseRubyEngineAndVersion(core.getInput('ruby-version'))
 
+    if (!platform.startsWith('windows') && version.match(/(mingw|mswin)$/)) {
+      throw new Error(`ruby-${version} is only available on windows`)
+    }
```
eregon: This shouldn't be necessary, getAvailableVersions + validateRubyEngineAndVersion will already check that below

windows.js
```diff
@@ -46,6 +48,25 @@ export async function install(platform, ruby) {
   return rubyPrefix
 }
 
+// all standard msvc OpenSSL builds use C:\Program Files\Common Files\SSL
+// as per openssl/openssl
+function mswinSSL(hostedRuby) {
```
eregon: Could we just set `SSL_CERT_FILE` instead like it's done for 2.2 and 2.3 above?

index.js
```diff
@@ -55,7 +59,9 @@ function parseRubyEngineAndVersion(rubyVersion) {
   } else { // engine-X.Y.Z
     [engine, version] = rubyVersion.split('-', 2)
   }
-
+  if (os.platform() === 'win32' && version === 'mingw') {
+    version = 'head'
+  }
```
eregon: I think a nicer way to do this would be to simply have both `head` and `mingw` in windows-versions.js (with the same URL).

windows-versions.js
```diff
@@ -24,5 +24,6 @@ export const versions = {
   "2.6.4": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.6.4-1/rubyinstaller-2.6.4-1-x64.7z",
   "2.6.5": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.6.5-1/rubyinstaller-2.6.5-1-x64.7z",
   "2.7.0": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.7.0-1/rubyinstaller-2.7.0-1-x64.7z",
-  "head": "https://github.com/oneclick/rubyinstaller2/releases/download/rubyinstaller-head/rubyinstaller-head-x64.7z"
+  "head": "https://github.com/MSP-Greg/ruby-loco/releases/download/ruby-master/ruby-mingw.7z",
```
eregon: I'm a bit hesitant to switch from rubyinstaller-head to ruby-loco here, because all release Windows builds are from RubyInstaller, so it creates some inconsistency.

Maybe we should allow both?
How about having `head` refer to rubyinstaller-head and mingw + mswin refer to ruby-loco builds?

eregon:
Fixes #22

index.js
MSP-Greg: That's fine.  I thought a specific message might be helpful, I'll remove.

.github/workflows/test.yml
MSP-Greg: Again, this is mess.  I've defined the string as an ENV variable in actions-ruby.

The `vcvars*.bat` files are needed when using msvc from the command line, as the scripts set up the environment.  So, if one needs the tools in multiple steps (which isolate the ENV from user script changes), one needs to run the script in each one until there's a better solution.  I am looking into one.

Some of this is made messier by Actions image options.  I haven't checked recently, but they announced that they were removing the Window-2016 image.  They are still updating the code for it.

So, although they've stated that they'll only have one Windows image, that will likely change when they release another version of Windows Server.  That's the image side, then there is the Visual Studio ('VS') side.  There's only one version of VS in each image, but that may also change.  So, again, I may see if I can get answers about the singular/multiple issues, as that effects the code and needed inputs in custom actions.

eregon: If https://github.com/MSP-Greg/actions-ruby/blob/164cd7048c050264a0360510c572e1d4472b4bfb/index.js#L139 is enough to let `gem install json` work with mswin builds, I think we should do that in this action too.

windows.js
MSP-Greg: With 2.2. and 2.3, the reason for using the ENV settings is that the OpenSSL package used pointed to a folder specific to the system the builder used.

The path for normal Windows OpenSSL builds is actually in the openssl/openssl code.  MSYS2 changes it to reference the location of the exe using OpenSSL.  Actions should have the files located there, but I don't want to assume that.

So, I'd prefer to set up OpenSSL as expected, and consider the 2.2 and 2.3 builds the exceptions.

index.js
MSP-Greg: Ok.  I wasn't sure what your thoughts about naming were.

I thought it would be helpful to have mingw master with 'ruby-head', 'head', and 'mingw'.  I'll remove the logic and add duplicate rows in windows-versions.js

windows-versions.js
MSP-Greg: Ok.

windows.js
eregon: I would think creating an empty `C:\\Program Files\\Common Files\\SSL\\certs` directory has no effect, so I would just skip and only check the cert.pem.
I'd prefer to simplify and just check `C:\\Program Files\\Common Files\\SSL\\cert.pem` exists and otherwise throw.

If the cert.pem file was missing, would the `ruby -ropen-uri -e 'puts open(%{https://rubygems.org/}) { |f| f.read(1024) }'` check file? If so, I think we can use that as our check of "OpenSSL works and can check certificates", without any extra code in windows.js.

.github/workflows/test.yml
MSP-Greg: Ok.  I'll set it up assuming 'singular' (as does actions-ruby) and hope we can keep an eye on Actions images.

windows.js
MSP-Greg: > If the cert.pem file was missing, would the ruby -ropen-uri -e 'puts open(%{https://rubygems.org/}) { |f| f.read(1024) }'

It fails.  Just tried locally after renaming my cert file in 'C:\Program Files\Common Files\SSL`.

> C:\\Program Files\\Common Files\\SSL\\certs

It is a correctly set up OpenSSL, and people may not check for its existence...

eregon: I think the check in the workflow is enough then.
If people rely on a directory being in the image, that's their responsibility.
`setup-ruby` should care about providing a working Ruby, including a working OpenSSL C extension, but not try to "lint" the whole system. There might be exceptions if we know something that is particularly unreliable, but for now I think the open-uri check above is good enough for testing OpenSSL.

MSP-Greg: My concern was more about having all the Ruby builds setup the same.  If some correctly connect a cert file, all should.  Same for the directory.  I'll check.

MSP-Greg: X509::DEFAULT_CERT_DIR exists on all builds, see:
https://github.com/MSP-Greg/ruby-setup-ruby-info/runs/460598415

head builds are failing, due to an issue with ruby/openssl.  Fixed there, not pushed to ruby/ruby yet.

windows-versions.js
MSP-Greg: Done.

MSP-Greg:
So I did a force push and nothing ran...  Not my day.

.github/workflows/test.yml
MSP-Greg: Done.  The Actions job runs it both in a cmd shell and the default PowerShell shell.  A little different.  I've got a bunch of issues to open re the Actions runner, one will be about adding the setting in VCVARS to the env with an action...

eregon:
> So I did a force push and nothing ran... Not my day.

I can trigger actions on `pull_request` if that helps.
Given it's triggering quite a lot of jobs I was a bit hesitant, since anyway they should run the fork's actions too on push.

.github/workflows/test.yml
```diff
@@ -9,42 +9,103 @@ on:
...
+    - run: bundle exec rake --version
+
+  specials:
+    strategy:
+      fail-fast: false
+      matrix:
+        cfg:
+          - { ruby: rubinius, os: ubuntu-18.04   }
+          - { ruby: mingw   , os: windows-latest }
+          - { ruby: mswin   , os: windows-latest }
```
eregon: Interesting syntax, that's cool :)
It's unfortunate we have to copy the steps, but I guess there is no other way.

windows.js
eregon: I see, it just fails if we don't setup those files: https://github.com/ruby/setup-ruby/runs/465758532

eregon:
@MSP-Greg I documented those in the README:
https://github.com/ruby/setup-ruby/commit/739b731933fe50ef039ae6ee56672f9042f4684c
Does that sound right?

eregon:
Released as https://github.com/ruby/setup-ruby/releases/tag/v1.20.0 and `v1` bumped.

MSP-Greg:
@eregon 

Thanks.  On to the package stuff.

___
[`mingw` and `mswin` are `ruby-head` builds using the MSYS2 and MSVC toolchains on Windows.]()
___

I think that's clear, but we (along with most people in the ruby org) know about them.  Maybe something like:

___
[On Windows, `ruby-head` & `mingw` are builds of master/head using MSYS2/MinGW gcc, which is the standard compiler used for Windows Rubies.  An `mwsin` build is also available, which is compiled with MSVC.  Both compilers are tested against in ruby/ruby.]()
___

It needs to be short, but hopefully it explains the builds for someone who knows nothing about Ruby on Windows...  I'll defer to your judgement.  

IMO, the people that need to know what `mswin` is already know.  One can't even use `mswin` with RuboCop, as jaro-winkler won't compile with it, and the current release doesn't force `mswin` to use the 'pure Ruby' implementation, it just fails...

MSP-Greg:
> Interesting syntax, that's cool :)

See https://github.com/actions/runner/issues/343

Officially supported and soon to be improved.

eregon:
@MSP-Greg Excellent! I wonder if we could merge again the test steps to avoid duplication with the new semantics for `include:`.

MSP-Greg:
> Should roll out to all accounts sometime over the next week or two.

It does take one to two weeks...

`ENV['ImageOS']` is now available on Ubuntu & Windows.  I think `os.platform` is still needed, but the combination of `ImageOS` and `ImageVersion` lets one determine exactly what 'virtual environment' image one is running on.  For instance, on windows-latest, `ImageOS` is `win19`, and on ubuntu-latest, `ImageOS` is `ubuntu18`.  On macOS, neither setting exists...

larskanis:
I didn't notice this change, but I like it! Now all interesting Windows Ruby versions are available through one action - great work!

PR #31, opened 2020-02-29
windows.js - use proper case for ENV['Path']

MSP-Greg:
Both locally and on Actions, the case used is 'Path'

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4929,22 +4929,23 @@ async function linkMSYS2() {
   return [latestHostedRuby, msys2]
 }
 
+// Current windows uses 'Path' for ENV['Path'], use proper case
 function setupPath(msys2, rubyPrefix) {
-  const originalPath = process.env['PATH'].split(';')
+  const originalPath = process.env['Path'].split(';')
   let path = originalPath.slice()
 
-  // Remove default Ruby in PATH
+  // Remove default Ruby in Path
   path = path.filter(e => !e.match(/\bRuby\b/))
 
   if (msys2) {
-    // Add MSYS2 in PATH
+    // Add MSYS2 in Path
     path.unshift(`${msys2}\\mingw64\\bin`, `${msys2}\\usr\\bin`)
   }
 
-  // Add the downloaded Ruby in PATH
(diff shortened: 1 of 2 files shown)


eregon:
Looks good. Is this the standard case on Windows? From what I remember env variables are case insensitive on Windows, right?

MSP-Greg:
> From what I remember env variables are case insensitive on Windows, right?

In theory.  But, one may have issues if one is dumping ENV to an object/hash.  Also made worse by the fact that Windows allows both '/' and '\\' for path separators...

I hate this.  I was just thinking of opening an issue in one of the GitHub Actions repos.

I also have a point of reference that is biased by the needs of building and testing ruby master, where having only one Ruby in PATH is helpful for CI.

So, I'm not sure whether this PR should be merged or whether we should remove the code below, which makes sure that only one Ruby install is in PATH:
```
core.exportVariable('Path', newPath)
```

I was working on setup-ruby-pkgs, and wondered if using `core.exportVariable('Path', str1)` and also `core.addPath(str2)`, both here and possibly in subsequent actions could be a little unstable?

MSP-Greg:
The above msg is a little general.

Things got a little squirrely when I had a `core.exportVariable('Path', str1)` command in setup-ruby-pkgs, which may have overwritten the `core.exportVariable` and/or `core.addPath` commands used in setup-ruby.

So, I think ``core.exportVariable('Path', str1)`` commands should not be used.  Thoughts?

JFYI, the issue that I had in ruby-setup-pkgs was a conflict with the multitude of build tools in 'Path' on the Actions Windows images.  I had to rename a 'gmake.exe' file...

MSP-Greg:
I'm working on another PR

eregon:
I'm OK to merge this FWIW

MSP-Greg:
I hope that PR https://github.com/ruby/setup-ruby/pull/32 supersedes this?

eregon:
Let's close this since #32 was merged.
We can use the proper name on Windows in `index.js`'s `setupPath()`, please make a new PR if you want that.

PR #32, opened 2020-02-29
reorganize Path handling, add MSVC env to action

MSP-Greg:
REVISED - 3 commits

1. Update package.json & yarn.lock
---
2. test.yml - Add which/where, mswin vcvars*.bat ENV
   * Add which/where for ruby and rake
   * Remove call %VCVARS% code for mswin
   * Loosen some step 'if' checks
---
3. Reorganize path manipulation, mswin tools ENV
   * Use only one 'core' call to set PATH
   * clean PATH
   * Windows mswin - add vcvars*.bat changes to ENV

### DIFF
--- a/ruby-builder.js
+++ b/ruby-builder.js
@@ -1,6 +1,5 @@
 const os = require('os')
 const path = require('path')
-const core = require('@actions/core')
 const exec = require('@actions/exec')
 const io = require('@actions/io')
 const tc = require('@actions/tool-cache')
@@ -15,17 +14,7 @@ export function getAvailableVersions(platform, engine) {
 
 export async function install(platform, ruby) {
   const rubyPrefix = await downloadAndExtract(platform, ruby)
-
-  if (platform === 'windows-latest') {
-    require('./windows').setupPath(undefined, rubyPrefix)
-  } else {
-    core.addPath(path.join(rubyPrefix, 'bin'))
-    if (ruby.startsWith('rubinius')) {
-      core.addPath(path.join(rubyPrefix, 'gems', 'bin'))
-    }
-  }
-
-  return rubyPrefix
+  return [rubyPrefix, null]
 }
 
 async function downloadAndExtract(platform, ruby) {
--- a/windows.js
+++ b/windows.js
@@ -2,6 +2,7 @@
 // https://github.com/MSP-Greg/actions-ruby/blob/master/lib/main.js
 
 const fs = require('fs')
+const cp = require('child_process')
 const core = require('@actions/core')
 const exec = require('@actions/exec')
 const tc = require('@actions/tool-cache')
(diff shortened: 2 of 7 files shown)


eregon:
Re 1, why is `core.exportVariable('PATH', newPath)` an issue?
That removes the system Ruby from PATH which seems a good idea.
We should maybe do the same on Linux/macOS too, although it seems less of a problem there.

The way `exportVariable` seems implemented is every following step gets a `PATH` entry in the `env:`, so other actions should just see this new path and compose nicely. What's the issue?

windows.js
```diff
@@ -48,64 +49,84 @@ export async function install(platform, ruby) {
...
+  const cert = 'C:\\Program Files\\Common Files\\SSL\\cert.pem'
+  if (!fs.existsSync(cert)) {
+    const hostedCert = `${hostedRuby}\\ssl\\cert.pem`
+    fs.copyFileSync(hostedCert, cert)
   }
 
-  // Add the downloaded Ruby in PATH
-  path.unshift(`${rubyPrefix}\\bin`)
+  addVCVARSEnv()
+  core.exportVariable('VCVARS', '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"')
```
eregon: VCVARS seems exported twice

MSP-Greg:
> why is core.exportVariable('PATH', newPath) an issue?

I was working with the mswin vcvars code in setup-ruby-pkgs, and I was also using `core.exportVariable('PATH', newPath)`.  It seemed that multiple calls did not work.  The last one overwrote the earlier ones.  IOW, each call starts with the system 'PATH', not the path created by all the previous `core.exportVariable('PATH', newPath)` calls.

This might be considered a bug in the core code, I didn't open an issue, as I was wondering whether a method like `core.removePath` (similar to `core.addPath`) could work...

eregon:
> It seemed that multiple calls did not work.

Do you have a log?
That seems a clear bug, obviously the new PATH should be used for every following step.

eregon:
FWIW I got very confusing results/errors when having the system Ruby in PATH and using JRuby on Windows, so I'm not so keen on keeping the system Ruby in PATH.

MSP-Greg:
> Do you have a log?

Maybe.  I was running code and running Actions for testing, then I removed the successive `core.exportVariable('PATH', newPath) calls...

> That seems a clear bug, obviously the new PATH should be used for every following step.

I'll open one and see what they say...

> FWIW I got very confusing results/errors when having the system Ruby in PATH and using JRuby on Windows, so I'm not so keen on keeping the system Ruby in PATH.

I was worried about that.  And as you stated, the results/errors are confusing.  It's the same when multiple build tools are in the path...

As you mentioned, it would be a good idea to remove the system Ruby from PATH in all OS's.

eregon:
> I'll open one and see what they say...

Thanks.
If you update this PR to still remove the system Ruby from the PATH on Windows, or on all platforms, I would be happy to merge it.

eregon:
The vcvars change is great BTW :)

MSP-Greg:
I added a commit to the PR.  Added 'setupPath' functions to ruby-builder.js & windows.js.  I set them up so just one 'core' call is used for the PATH manipulation, and also kept the cleaned path and the additions separate until the end.  The code is relatively clean, except for the JRuby on Windows 'exception'.

I added which/where steps for 'ruby' and 'rake' to the CI.  Windows is clean, but please look at the Ubuntu & macOS, as I'm not sure what the best way to 'remove' the system Ruby is.

Anyway, not sure what you might want for re-organizing the commits, etc...

MSP-Greg:
I added path cleaning on macOS, there's nothing to clean on Ubuntu.

MSP-Greg:
Reorganized commits...

ruby-builder.js
```diff
@@ -15,17 +16,40 @@ export function getAvailableVersions(platform, engine) {
 
 export async function install(platform, ruby) {
   const rubyPrefix = await downloadAndExtract(platform, ruby)
+  return rubyPrefix
+}
+
+export async function setupPath(rubyPrefix, ruby) {
+  const rubyPath = path.join(rubyPrefix, 'bin', 'ruby')
+  const gemUserDir = cp.execSync(`${rubyPath} -e "puts Gem.user_dir"`).toString().trim()
```
eregon: Why do you add the Gem.user_dir here?
AFAIK nobody bothers with `gem install --user-dir` in CI, they expect they can just `gem/bundle install`.

ruby-builder.js
```diff
@@ -15,17 +16,40 @@ export function getAvailableVersions(platform, engine) {
...
 
-  if (platform === 'windows-latest') {
-    require('./windows').setupPath(undefined, rubyPrefix)
+  const originalPath = process.env['PATH'].split(path.delimiter)
+
+  // macOS & Windows have ruby folders in PATH, Ubuntu does not
```
eregon: It would be more reliable to always remove any path in PATH which matches `/\bruby\b/i`, no matter the platform.

eregon: It would be nice to share the common part of this code between the two installers to avoid duplication.
I think Windows just needs to add a few extra entries.

eregon: It looks like there is `/usr/bin/ruby` in PATH though on Ubuntu and macOS, and we shouldn't remove `/usr/bin` from PATH of course.

MSP-Greg: > It would be nice to share the common part

I was debating that also.   I'll see what I can do.

ruby-builder.js
MSP-Greg: > Why do you add the Gem.user_dir here?

The `"#{Gem.user_dir}/bin"` directory.  Just because someone, someday, will ask about it, and it's not hard to include...

eregon: I prefer to keep the logic simpler here and not try to guess what someone, someday, will ask.

ruby-builder.js
MSP-Greg: > It looks like there is /usr/bin/ruby in PATH though on Ubuntu and macOS

Maybe you're looking at the which output?

See [actions-image-testing](https://github.com/MSP-Greg/actions-image-testing/runs/491413875?check_suite_focus=true#step:4:1)

eregon: Yes, I was looking at the checks on this PR, e.g. https://github.com/ruby/setup-ruby/pull/32/checks?check_run_id=491284376

ruby-builder.js
eregon: Also shelling out to `ruby` should be avoided, that will slow down every usage of the action. On JRuby it can take easily 1 second just to shell out to print `Gem.user_dir`.

MSP-Greg: See https://github.com/rubygems/rubygems/pull/2847

Not implemented yet, but...

eregon: Thanks for the link. I don't like to speculate on the future though, and I want to keep this action as simple as possible, simply because otherwise I won't have time to maintain it.

ruby-builder.js
MSP-Greg: > Yes, I was looking at the checks on this PR

I guess without a hard link to the line, I don't know.  But the job log outputs `/usr/bin/ruby` as the output of a `which -a ruby` command...

Do you want the two lines of code that determine the `--user-install` bin folder commented out or removed?

MSP-Greg:
Revised...

ruby-builder.js
eregon: Removed, maintaining commented code is no fun.

eregon:
Released as https://github.com/ruby/setup-ruby/releases/tag/v1.22.0

PR #34, opened 2020-03-05
Add JRuby 9.2.11.0

deivid-rodriguez:
Not exactly sure what the process for adding new rubies is, I just "copied" 7f508b5101e9cc2d7cbb7851689507771d9ff6c6.


deivid-rodriguez:
I see by the failures that this building this ruby needs to be triggered by a push to https://github.com/ruby/ruby-builder, right?

MSP-Greg:
Yes.  This repo just handles the download & installing in the Action runner.

As you mentioned, ruby-builder is where all the binaries are built and stored, except for head/master builds & Windows MRI builds...  (I think)

eregon:
Yep, that's documented in https://github.com/ruby/setup-ruby#supported-platforms.
I'll trigger the build in ruby-builder.

eregon:
Merged, released, v1 updated.

deivid-rodriguez:
Thanks!

PR #35, opened 2020-03-09
Add JRuby 9.1.17.0 for testing

headius:
The jruby-launcher needs to be tested at least one major JRuby version back. In addition, there are users out there running JRuby 9.1.x that may be unable to upgrade at present.

### DIFF
--- a/ruby-builder-versions.js
+++ b/ruby-builder-versions.js
@@ -10,7 +10,7 @@ export function getVersions(platform) {
       "head"
     ],
     "jruby": [
-      "9.2.9.0", "9.2.10.0", "9.2.11.0",
+      "9.1.17.0", "9.2.9.0", "9.2.10.0", "9.2.11.0",
       "head"
     ],
     "truffleruby": [


ruby-builder-versions.js
```diff
@@ -10,7 +10,7 @@ export function getVersions(platform) {
       "head"
     ],
     "jruby": [
-      "9.2.9.0", "9.2.10.0", "9.2.11.0",
+      "9.1.17.0", "9.2.9.0", "9.2.10.0", "9.2.11.0",
```
eregon: Could you add `9.1.17.0` on its own line, much like 2.4.x vs 2.5.x in MRI above?

eregon:
OK, I wasn't sure projects test against older JRuby versions or if they just use 9.2.x.y.

Could you also update the README, and regenerate `dist/index.js` with `yarn run package` ?
Like in https://github.com/ruby/setup-ruby/pull/34/files
(I should add that in CONTRIBUTING.md)

deivid-rodriguez:
I was about to request this too. Since you still provide the legacy ruby 2.3 option, it makes sense to provide jruby-9.1 too since it's MRI 2.3 compatible.

eregon:
I thought JRuby 9.2.x would be compatible with 2.4.x but actually it's 2.3.x (MRI 2.3.x is already EOL).
I had to add MRI 2.3.x to ease migration from TravisCI, so OK to add JRuby 9.2.x too, especially since it seems easy to build/setup.

eregon:
CONTRIBUTING.md added: https://github.com/ruby/setup-ruby/blob/master/CONTRIBUTING.md

headius:
> I thought JRuby 9.2.x would be compatible

> OK to add JRuby 9.2.x too

I assume you mean 9.1.x here.

eregon:
(yes)

headius:
Not sure what else I am supposed to install to be able to run `yarn run package`:

```
$ yarn run package
yarn run v1.22.4
$ ncc build index.js -o dist
/bin/sh: ncc: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

headius:
Is it this?

https://www.npmjs.com/package/@zeit/ncc

eregon:
Did you run `yarn` before to install the packages?
I don't have `ncc` installed system-wide and it works fine:
```
$ yarn run package
yarn run v1.17.3
$ ncc build index.js -o dist
ncc: Version 0.21.1
ncc: Compiling file index.js
164kB  dist/index.js
164kB  [827ms] - ncc 0.21.1
Done in 1.03s.

$ find . -name ncc
./node_modules/@zeit/ncc
./node_modules/@zeit/ncc/dist/ncc
./node_modules/.bin/ncc
```

eregon:
i.e., please read the new CONTRIBUTING.md :)

headius:
> i.e., please read the new CONTRIBUTING.md :)

Not sure why you would include this comment.

I did read that document but it wasn't clear from the formatting that I was supposed to **run** yarn rather than just installing it. It's just the word `yarn` in the document under the heading "Install dependencies". Perhaps it would be clearer to include text indicating that's a command to be run rather than just a list of dependencies.

headius:
Or include output showing a command prompt, so it's clear what are commands.

headius:
Running `yarn run package` creates a very large diff that I don't want to push. Is this expected?

https://gist.github.com/headius/f3e919c4331ed5a92bf56cefd0a2c80c

headius:
https://github.com/ruby/setup-ruby/pull/37

eregon:
@headius That seems because you edited `package.json`: https://gist.github.com/headius/f3e919c4331ed5a92bf56cefd0a2c80c#file-diff-diff-L2412

eregon:
> Or include output showing a command prompt, so it's clear what are commands.

OK, added that in 166761b3102be072704f1cc99974249ffe8e8c5a

eregon:
Released: https://github.com/ruby/setup-ruby/releases/tag/v1.23.0

headius:
> That seems because you edited package.json

I would not have touched that file... it must have been yarn. I only edited ruby-builder-versions.js and then ran the commands you specified in CONTRIBUTING.md. Looking at it today, seems like it's because the existing files were generated with an older version of yarn and/or ncc.

PR #39, opened 2020-03-24
Update for Actions Windows MSYS2 install

MSP-Greg:
1. Use 'Git for Windows' cert.pem file
2. Add MSYS2 to path for mswin (after msvc additions)
3. Rearrange code for cleanup when MSYS2 is added to Window image

The comment below precedes code that can be removed when MSYS2 is added.
```
// Remove when Actions Windows image contains MSYS2 install
```

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -8,6 +8,12 @@ const exec = require('@actions/exec')
 const tc = require('@actions/tool-cache')
 const rubyInstallerVersions = require('./windows-versions').versions
 
+// needed for 2.2, 2.3, and mswin, cert file used by Git for Windows
+const certFile = 'C:\\Program Files\\Git\\mingw64\\ssl\\cert.pem'
+
+// standard MSYS2 location, found by 'devkit'
+const msys2 = 'C:\\msys64'
+
 export function getAvailableVersions(platform, engine) {
   if (engine === 'ruby') {
     return Object.keys(rubyInstallerVersions)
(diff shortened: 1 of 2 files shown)


windows.js
```diff
@@ -58,36 +62,37 @@ function latestHostedRuby() {
...
-    fs.copyFileSync(hostedCert, cert)
+    fs.copyFileSync(certFile, cert)
   }
-
-  return addVCVARSEnv()
+  let pathAry = addVCVARSEnv()
+  // add MSYS2 paths for misc gnu utilities like bison and ragel
+  pathAry.push(`${msys2}\\mingw64\\bin`)
+  pathAry.push(`${msys2}\\usr\\bin`)
+  return pathAry
```
eregon: Could you use an array literal with `...` to splat addVCVARSEnv()?
That would be closer to the code above in `setupMingw` and avoid mutating the array in place.

windows.js
```diff
@@ -8,6 +8,12 @@ const exec = require('@actions/exec')
 const tc = require('@actions/tool-cache')
 const rubyInstallerVersions = require('./windows-versions').versions
 
+// needed for 2.2, 2.3, and mswin, cert file used by Git for Windows
+const certFile = 'C:\\Program Files\\Git\\mingw64\\ssl\\cert.pem'
```
eregon: I hope this location will remain available, it sounds likely enough for now.

MSP-Greg: Git for Windows has always been there.  It's the git that's used by all the actions and any Windows devs using git.  There are also GitHub/MSFT staff involved with it.  IOW, its location is probably more locked down than the locations of the toolcache Rubies...

windows.js
MSP-Greg: I noticed that push allows multiple arguments, so I changed to that?   VCVARSEnv is defined by the bat file, so we have to 'add' two arrays, but this is just adding two items...

eregon: I meant https://github.com/ruby/setup-ruby/commit/ae39a8d7d26f4a326f81694e5e99ed009d312fdf actually

eregon:
Released as https://github.com/ruby/setup-ruby/releases/tag/v1.24.0

PR #41, opened 2020-03-27
Windows 2.2 & 2.3 - install old MSYS DevKit

MSP-Greg:
Three commits:

1. 'windows.js - install old MSYS devkit for MinGW 2.2 and 2.3' - adds code to install MSYS DevKit for Windows MinGW  2.2 and 2.3 Rubies

2. 'Actions - add gcc --version output for MinGW Rubies' - added code to show version of gcc in Path.  2.2 and 2.3 should show 4.7.2 gcc, 2.4 and later should show 8.3.0 or later

3. 'README.md - add info re MSYS install for MinGW 2.2 and 2.3 Rubies'

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,14 +33,17 @@ jobs:
     - name: build compiler
       run: |
         ruby -e "puts 'build compiler: ' + RbConfig::CONFIG.fetch('CC_VERSION_MESSAGE', 'unknown').lines.first"
-    - name: ridk version (mingw)
+    - name: gcc and ridk version (mingw)
       if: startsWith(matrix.os, 'windows')
       run: |
         $abi, $plat = $(ruby -e "STDOUT.write RbConfig::CONFIG['ruby_version'] + ' ' + RUBY_PLATFORM").split(' ')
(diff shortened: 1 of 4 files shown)


eregon:
Could you do the `export` PR? Then I'll make a release.

eregon:
Released: https://github.com/ruby/setup-ruby/releases/tag/v1.25.0

MSP-Greg:
@eregon

I came across something interesting while revising code in pkgs.  Previously I had the old DevKit code in it, and Puma was using it, so I had test logs in both.

When I moved the code here in the PR, I changed the drive from D: (ssd) to C: (hd), kind of thinking that the MSYS2 install would be there, might as well put the older 'DevKit' there.

The DevKit has a lot of files, I think around 7k.  From the logs, extracting it to the C: drive takes about twice as long as the SSD drive (from ~10 to ~20 sec).  I looked at several logs because the difference really surprised me.

So, any objections to moving the location to D:?  This needs to stay in sync with pkgs, as it extracts the few packages that are available for old Rubies into it...

eregon:
:+1: for faster setup of MSYS

PR #45, opened 2020-03-30
Windows - set ENV['MAKE'] to avoid conflicts with other Actions tools

MSP-Greg:
To avoid things like:
```yml
    env:
      MAKE: "make" # to not try to use gmake on Windows
```

I had similar code in pkgs.  Should be here.  I also checked, and this is the last of things affecting 'build tools' (that have been noticed)...

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4879,6 +4879,8 @@ async function symLinkToEmbeddedMSYS2() {
 }
 
 async function setupMingw(version) {
+  core.exportVariable('MAKE', 'make.exe')
+
   if (version.startsWith('2.2') || version.startsWith('2.3')) {
     core.exportVariable('SSL_CERT_FILE', certFile)
     await installMSYS(version)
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Ok, mistaken.  AFAIK, the Windows & Ubuntu CI runs on the same type of Azure machine.  Windows has always had two native ENV variables, TEMP & TMP, both of which are always defined on the drive that the OS is on.  The Ubuntu CI does not define either.

In my general (all platforms) code, I set:
```js
core.exportVariable('TMPDIR', process.env.RUNNER_TEMP)
```
Which is on the SSD.  I believe it's the first ENV variable that Ruby checks for temporary locations (dir/files).

I can add it, we can forget it, or you can add it...

eregon:
You mean https://github.com/ffi/ffi/commit/6cc11cf137e6dcce00b5740e465f6ff3ccb69662, right?

Yes, great PR, let's merge it.

eregon:
Cleanup in FFI: https://github.com/ffi/ffi/commit/e116eb8f0fcd5fae5a6566955326a0f2bca3ca16

PR #48, opened 2020-04-02
gem install bundler - add '--conservative', small README.md change

MSP-Greg:
Two commits:

1. 'gem install bundler - add `--conservative`' - to avoid unnecessary bundler installs, especially on master

2. `README.md - remove reference to MSP-Greg/actions-ruby` - Preferred repo is now MSP-Greg/setup-ruby-pkgs

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,5 +163,5 @@ Please [update](https://github.com/ruby/setup-ruby/releases/tag/v1.13.0) if you 
 ## Credits
 
 The current maintainer of this action is @eregon.
-Most of the Windows logic is from https://github.com/MSP-Greg/actions-ruby by MSP-Greg.
+Most of the Windows logic is based on work by MSP-Greg.
 Many thanks to MSP-Greg and Lars Kanis for the help with Ruby Installer.
--- a/dist/index.js
+++ b/dist/index.js
@@ -1107,7 +1107,7 @@ async function installBundler(platform, rubyPrefix, engine, rubyVersion) {
   } else if (/^\d+/.test(bundlerVersion)) {
     versionArray = ['-v', `~> ${bundlerVersion}`]
   } else if (bundlerVersion === 'latest') {
-    versionArray = []
+    versionArray = ['-v', '>= 2.1']
   } else {
     throw new Error(`Cannot parse bundler input: ${bundlerVersion}`)
   }
@@ -1117,7 +1117,7 @@ async function installBundler(platform, rubyPrefix, engine, rubyVersion) {
   } else if (bundlerVersion === '1' && engine === 'truffleruby') {
     console.log(`Using the Bundler version shipped with ${engine}`)
   } else {
-    await exec.exec(path.join(rubyPrefix, 'bin', 'gem'), ['install', 'bundler', ...versionArray, '--no-document'])
+    await exec.exec(path.join(rubyPrefix, 'bin', 'gem'), ['install', 'bundler', ...versionArray, '--no-document', '--conservative'])
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -149,7 +149,7 @@ async function installBundler(platform, rubyPrefix, engine, rubyVersion) {
   } else if (/^\d+/.test(bundlerVersion)) {
     versionArray = ['-v', `~> ${bundlerVersion}`]
   } else if (bundlerVersion === 'latest') {
-    versionArray = []
+    versionArray = ['-v', '>= 2.1']
   } else {
     throw new Error(`Cannot parse bundler input: ${bundlerVersion}`)
   }
@@ -159,7 +159,7 @@ async function installBundler(platform, rubyPrefix, engine, rubyVersion) {
   } else if (bundlerVersion === '1' && engine === 'truffleruby') {
     console.log(`Using the Bundler version shipped with ${engine}`)
   } else {
-    await exec.exec(path.join(rubyPrefix, 'bin', 'gem'), ['install', 'bundler', ...versionArray, '--no-document'])
+    await exec.exec(path.join(rubyPrefix, 'bin', 'gem'), ['install', 'bundler', ...versionArray, '--no-document', '--conservative'])
   }
 }
 


MSP-Greg:
Working with master, I often get frustrated with RG/Bundler.  This is not working as expected.  In my fork, I looked at the head builds, and they're not installing Bundler (as desired).

But, if you compare the 2.5 & 2.6.5 jobs, in the ' Run /./' and 'Run bundle --version' steps, it certainly looks like something isn't working right.  2.5 seems right, but 2.6.5 is a 'WTF is this?'...

Maybe use a regex like `rubyVersion.match(/(head|mingw|mswin)$/)` and skip `gem install bundler` for them?

eregon:
> Working with master, I often get frustrated with RG/Bundler. This is not working as expected. In my fork, I looked at the head builds, and they're not installing Bundler (as desired).

I'm not sure what you mean, could you point to a log and explain your expectation?
My expectation is it should install latest bundler on ruby-head, and basically every Ruby >= 2.3 if there is no Gemfile.lock by default.

> but 2.6.5 is a 'WTF is this?'...

Log please :)

MSP-Greg:
> My expectation is it should install latest bundler on ruby-head

Right now, head is running with RubyGems 3.2.0.pre1, which isn't released.  The same may be true for Bundler.  Hence, ruby-head should not have anything done to its RubyGems or Bundler.  If a Gemfile, etc requires Bundler 1.x, then it should be installed.

> Log please :)

The log for the CI on this PR.

Ruby 2.5
```
Step:  Run /./
  Successfully installed bundler-2.1.4

Step:  Run bundle --version
  Bundler version 2.1.4
```
Ruby 2.6.5
```
Step:  Run /./
  No gem (Bundler) installed

Step:  Run bundle --version
  Bundler version 1.17.2
```

Why does 2.6.5 have Bundler 1.17.2, but 2.5 has 2.1.4?

MSP-Greg:
Fixed issue with Ruby 2.6, apparently RubyGems 3.0.3 needs a version requirement for --conservative to function as in other versions.  Ruby 2.6 is now correctly updating Bundler to the current version.

eregon:
https://github.com/ruby/setup-ruby/pull/48/checks?check_run_id=556815455 shows Bundler 2.1.4 for Ruby 2.6.5.
https://github.com/ruby/setup-ruby/pull/48/checks?check_run_id=556815516 too
and https://github.com/ruby/setup-ruby/pull/48/checks?check_run_id=556815580 too

So please, I need a log file where you see the issue :)

eregon:
Same results on the original PR: https://github.com/ruby/setup-ruby/runs/550219139?check_suite_focus=true (Bundler 2 on Ruby 2.6.5 on all 3 platforms)

eregon:
> Right now, head is running with RubyGems 3.2.0.pre1, which isn't released. The same may be true for Bundler. Hence, ruby-head should not have anything done to its RubyGems or Bundler. If a Gemfile, etc requires Bundler 1.x, then it should be installed.

That's something to consider. I think there might be cases where ruby-head has a older Bundler than the latest release. Bundler is updated infrequently in ruby-head IIRC.
I agree for ruby-head (and mingw/msiwn) it seems safer to not touch Bundler if the logic would otherwise decide to install the latest Bundler. In other words, for ruby-head, only install Bundler if we decide we need Bundler 1.

Regarding `--conservative`, would that use the Bundler 2.1.4 shipped with Ruby 2.7.1, instead of a more recent Bundler release, say 2.2.0?
I think that would be suboptimal if we have Ruby 2.6.5 gets the latest Bundler 2.2.0 (because it doesn't ship with Bundler 2) but Ruby 2.7.1 ends up with Bundler 2.1.4 (which might have bugs fixed in 2.2.0).

eregon:
Found https://github.com/ruby/setup-ruby/runs/556397578?check_suite_focus=true which indeeds reports Bundler 1 since `--conservative` added in this PR has no version requirement and so is happy with the default Bundler gem (which is Bundler 1).

eregon:
> But, if you compare the 2.5 & 2.6.5 jobs, in the ' Run /./' and 'Run bundle --version' steps, it certainly looks like something isn't working right. 2.5 seems right, but 2.6.5 is a 'WTF is this?'...

It was really not clear to me you were talking about the CI of this PR, I thought you were mentioning a bug of current `master`.

MSP-Greg:
Sorry.  I was irritated about RubyGems, as most versions seem to take `--conservative` as equivalent to 'don't re-install if on latest', but with RG 3.0.3, it's equivalent to 'don't install if it is installed'.

Either way, probably best to include the `>= 2.1`.

>  I thought you were mentioning a bug of current master.

I would have made that clear.  We got messed up with 'logs', and I did say "but 2.6.5 is a 'WTF is this?".  Or, your code is fine, it's the RG/Bundler quirks.  I wrestled with them with AppVeyor & Travis, I just want to be done messing with them.  JFYI, I have come across CI where RG needed an update when using Ruby 2.2...

MSP-Greg:
> Regarding --conservative, would that use the Bundler 2.1.4 shipped with Ruby 2.7.1, instead of a more recent Bundler release, say 2.2.0?
I think that would be suboptimal if we have Ruby 2.6.5 gets the latest Bundler 2.2.0 (because it doesn't ship with Bundler 2) but Ruby 2.7.1 ends up with Bundler 2.1.4 (which might have bugs fixed in 2.2.0).

I think we may need to wait and see how the RG in those versions handles `--conservative`?

Maybe just add conditional logic to leave head's Bundler alone, and remove `--conservative`?

eregon:
Any reason to use `>= 2.1` over just `~> 2`?
I guess that wouldn't allow Bundler 3, but I think we can revisit the situation then.

> Maybe just add conditional logic to leave head's Bundler alone, and remove `--conservative`?

Yes, I think we should do that, it seems safer to me and more predictable than using a flag which apparently changes meaning in different RubyGems versions.
I'll take care of that change, and merge your small README fix.

eregon:
`--conservative` would have the advantage to not reinstall Bundler 2.1.4 on Ruby 2.7.1 currently, without needing explicit logic.
OTOH then we could have the situation where there is a future Bundler 2.2.0 and it would be used for all Rubies (2.3/2.4/2.5/2.6) except head (kind of OK) and 2.7.1 (kind of surprising I'd say).

So I think for 2.7.1 it's not good to skip installing Bundler right now, since it would only be without side-effects until the next Bundler version comes out.
I might reconsider later, I'm not sure what's better here:
* use the stdlib Bundler if same major version as Gemfile.lock or
* try to use latest Bundler of the same major version so all or almost all Rubies use that same latest Bundler of the same major version (current approach)

eregon:
See c40ce85201dc1277b424f20b6960a18b54277520 and the parent commit, that will avoid installing Bundler on ruby-head/mingw/mswin.

I think the main issue is addressed here, thanks for reporting it and the PR.
Let's close this PR since it will conflict.

I'll keep thinking about the two approaches mentioned above.

PR #50, opened 2020-04-18
Add support for setting a working directory

FineTralfazz:
This commit adds a `working-directory` option. This is useful for cases where files like `.ruby-version` and `Gemfile.lock` aren't located in the root directory. In my case, this is because I'm working in a monorepo, where my Ruby project is located in its own subfolder.


eregon:
Released as https://github.com/ruby/setup-ruby/releases/tag/v1.32.0

PR #53, opened 2020-05-01
Use the `--deployment` flag for bundle install

scottjacobsen:
`bundle install --deployment` should be used on CI environments - it sets additional options beyond the install path (ensures the Gemfile.lock is checked in, up to date, etc). https://bundler.io/v2.0/man/bundle-install.1.html#DEPLOYMENT-MODE

`bundle --retry 3` is the default and does not need to be specified. https://bundler.io/v2.0/man/bundle-config.1.html#LIST-OF-AVAILABLE-KEYS

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,7 @@ You can cache the installed gems with these two steps:
         restore-keys: |
           bundle-use-ruby-${{ matrix.os }}-${{ matrix.ruby }}-
     - name: bundle install
-      run: |
-        bundle config path vendor/bundle
-        bundle install --jobs 4 --retry 3
+      run: bundle install --deployment --jobs 4
 ```
 
 When using a single OS, replace `${{ matrix.os }}` with the OS.  
--- a/README.md
+++ b/README.md
@@ -120,7 +120,9 @@ You can cache the installed gems with these two steps:
         restore-keys: |
           bundle-use-ruby-${{ matrix.os }}-${{ matrix.ruby }}-
     - name: bundle install
-      run: bundle install --deployment --jobs 4
+      run: |
+        bundle config deployment true
+        bundle install --jobs 4
 ```
 
 When using a single OS, replace `${{ matrix.os }}` with the OS.  


eregon:
Thanks for the PR!
`--deployment` needs a `Gemfile.lock` but for example some gems (e.g., https://github.com/ruby/rake) using this action in CI might not have a checked-in `Gemfile.lock`.
Do you have an idea how to address those 2 cases?

scottjacobsen:
Good point. I didn't think about gem development.

Maybe something like this?

It doesn't hurt to use `bundle config path` first and then `--deployment`

```yaml
run: |
        bundle config path vendor/bundle
        test -f Gemfile.lock && bundle install --deployment || bundle install
```

scottjacobsen:
I suppose there is some poor soul using windows and I assume the `test` command does not exist there. But I don't know. 

Maybe there should be separate examples for Gem devs and app devs. 

I also noticed a few lines earlier the `Gemfile.lock` is being hashed into the cache key - so the example already assumes there is a lock file.

eregon:
> I also noticed a few lines earlier the `Gemfile.lock` is being hashed into the cache key - so the example already assumes there is a lock file.

That's a good point, and probably for the gem development case without a `Gemfile.lock` it doesn't make much sense to cache gems, since the purpose is to test the newest dependencies.

README.md
```diff
@@ -120,9 +120,7 @@ You can cache the installed gems with these two steps:
         restore-keys: |
           bundle-use-ruby-${{ matrix.os }}-${{ matrix.ruby }}-
     - name: bundle install
-      run: |
-        bundle config path vendor/bundle
-        bundle install --jobs 4 --retry 3
+      run: bundle install --deployment --jobs 4
```
eregon, line 123: On Bundler 2 this triggers a warning:

$ bundle install --deployment
[DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set deployment 'true'`, and stop using this flag


Could you change it accordingly?

eregon: I'd suggest the shorter form `bundle config deployment true`.
That's global, but probably good in CI?
`bundle config --local deployment true` seems fine too.

scottjacobsen:
I used the `bundle config deployment true` setting. I tried to test it on my project and it seemed to work,but it threw up a different warning that my version of rails (4.2) depends on bundler < 2.0. So it isn't clear to me if it ran using bundler 2 or bundler 1.7. But it seemed to do the right thing.

sonicdoe:
Looking into the deprecation of `--deployment`, Bundler doesn’t (yet) list it as being deprecated [in their docs](https://bundler.io/bundle_install.html). [In their `UPGRADING.md`](https://github.com/rubygems/rubygems/blob/2d989434dfbf769cb42110cc3df767195b87813e/bundler/UPGRADING.md#cli-deprecations), though, they go into more detail and explain that these deprecations are printed since Bundler 2.1 and that the flags will probably be removed in Bundler 3.

However, an important caveat is that `bundle config set deployment true` didn’t actually behave in the same way as `--deployment` up until Bundler 2.1.3 (which includes https://github.com/rubygems/bundler/pull/7519). Most notably, Bundler 2.1.2 and earlier (including 1.17.3) don’t actually install gems to `vendor/bundle` if you activate deployment mode through `bundle config`.

Therefore, we might want to set both `deployment` _and_ `path` through `bundle config`.

eregon:
@sonicdoe Thanks for all the details.
So I guess the safest is to use both `bundle config path vendor/bundle` and `bundle config deployment true` then.
I'll merge this PR and adapt the README to mention that.

eregon:
@scottjacobsen What's the reason for wanting `--deployment`?

In some cases like #136 it's actually problematic, so I wonder what it gives.
It seems only point 2 of the [docs](https://bundler.io/v2.0/man/bundle-install.1.html#DEPLOYMENT-MODE) matters (Checks that the Gemfile.lock is up to date with the Gemfile).
Point 1 is not checked in this case, and point 3 is already done separately.

scottjacobsen:
Point 2 is what I was mostly worried about. I think the traditional way to handle that is with `--deployment`, and probably most CI services recommend using this flag. It will cause `bundle install` to fail if the `Gemfile` is out of sync with the `Gemfile.lock`.  

Also I believe installing to `vendor/bundle` is preferable because that is the closest to the production setup, and I would bet in most CI environments the locations you can write to are limited.

I skimmed issue #136 and I believe the best solution there is to configure any linters, etc to ignore vendor/bundle, as opposed to complicating this tool.

I'm not a total expert on bundler configuration - but can environment variables be used to override the options set by the `--deployment` option? If so users could set those values to customize their settings.

eregon:
Thanks for the reply, actually I mislinked, I meant #153 primarily.
`deployment` as the config level is actually separate from the `path` (the action uses `vendor/bundle` too for the path).

PR #56, opened 2020-05-16
Automated bundle install and caching

### LINKED ISSUE #52: Automated Bundler caching
I'd like to add automated Bundler caching in this action, since caching manually like https://github.com/ruby/setup-ruby/blob/master/README.md#caching-bundle-install is quite verbose and hard to get right (e.g., https://github.com/sass/sassc-ruby/issues/183#issuecomment-622357550).

It would be optional, not sure about the default yet.

For that to work, we need `actions/cache` to be a package, so this needs https://github.com/actions/cache/issues/55 first.

eregon:
See #52

- [ ] test what happens for cron actions

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -13,6 +13,10 @@ inputs:
     description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. The default tries Gemfile.lock and otherwise uses latest.'
     required: false
     default: 'default'
+  bundler-cache:
+    description: 'Run "bundle install", and cache the result automatically. Either true or false.'
+    required: false
+    default: 'false'
   working-directory:
     description: 'The working directory to use for resolving paths for .ruby-version, .tool-versions and Gemfile.lock.'
     required: false
--- a/index.js
+++ b/index.js
@@ -3,11 +3,13 @@ const fs = require('fs')
 const path = require('path')
 const core = require('@actions/core')
 const exec = require('@actions/exec')
+const cache = require('@actions/cache')
 const common = require('./common')
 
 const inputDefaults = {
   'ruby-version': 'default',
   'bundler': 'default',
+  'bundler-cache': 'true',
   'working-directory': '.',
 }
 
@@ -50,6 +52,11 @@ export async function setupRuby(options) {
   if (inputs['bundler'] !== 'none') {
     await common.measure('Installing Bundler', async () =>
       installBundler(inputs['bundler'], platform, rubyPrefix, engine, version))
+
+    if (inputs['bundler-cache'] === 'true') {
+      await common.measure('bundle install', async () =>
+          bundleInstall(platform, engine, version))
+    }
   }
 
   core.setOutput('ruby-prefix', rubyPrefix)
@@ -192,6 +199,83 @@ async function installBundler(bundlerVersionInput, platform, rubyPrefix, engine,
     const gem = path.join(rubyPrefix, 'bin', 'gem')
     await exec.exec(gem, ['install', 'bundler', '-v', `~> ${bundlerVersion}`, '--no-document'])
   }
+}
+
+async function bundleInstall(platform, engine, version) {
+  if (!fs.existsSync('Gemfile')) {
+    console.log('No Gemfile, skipping "bundle install" and caching')
+    return
+  }
+
+  // config
+  const path = 'vendor/bundle'
+  const hasGemfileLock = fs.existsSync('Gemfile.lock');
+  if (hasGemfileLock) {
+    await exec.exec('bundle', ['config', '--local', 'deployment', 'true'])
+  }
+  await exec.exec('bundle', ['config', '--local', 'path', path])
+
+  // cache key
+  const paths = [path]
+  const baseKey = await computeBaseKey(platform, engine, version)
+  let key = baseKey
+  let restoreKeys
+  if (hasGemfileLock) {
+    key += `-Gemfile.lock-${await common.hashFile('Gemfile.lock')}`
+    // If only Gemfile.lock we can reuse some of the cache (but it will keep old gem versions in the cache)
+    restoreKeys = [`${baseKey}-Gemfile.lock-`]
+  } else {
+    // Only exact key, to never mix native gems of different platforms or Ruby versions
+    restoreKeys = []
+  }
+  console.log(`Cache key: ${key}`)
+
+  // restore cache & install
+  const cachedKey = await cache.restoreCache(paths, key, restoreKeys)
+  if (cachedKey) {
+    console.log(`Found cache for key: ${cachedKey}`)
+  }
+
+  let alreadyInstalled = false
+  if (cachedKey === key) {
+    const exitCode = await exec.exec('bundle', ['check'], { ignoreReturnCode: true })
+    alreadyInstalled = (exitCode === 0)
+  }
+
+  if (!alreadyInstalled) {
+    await exec.exec('bundle', ['install', '--jobs', '4'])
+
+    // Error handling from https://github.com/actions/cache/blob/master/src/save.ts
+    console.log('Saving cache')
+    try {
+      await cache.saveCache(paths, key)
+    } catch (error) {
+      if (error.name === cache.ValidationError.name) {
+        throw error;
+      } else if (error.name === cache.ReserveCacheError.name) {
+        core.info(error.message);
+      } else {
+        core.info(`[warning]${error.message}`)
+      }
+    }
+  }
+}
+
+async function computeBaseKey(platform, engine, version) {
+  let baseKey = `setup-ruby-bundle-install-${platform}-${engine}-${version}`
+  if (engine === 'ruby' && common.isHeadVersion(version)) {
+    let revision = '';
+    await exec.exec('ruby', ['-e', 'print RUBY_REVISION'], {
+      silent: true,
+      listeners: {
+        stdout: (data) => {
+          revision += data.toString();
+        }
+      }
+    });
+    baseKey += `-revision-${revision}`
+  }
+  return baseKey
 }
 
 if (__filename.endsWith('index.js')) { run() }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,7 @@ jobs:
     - uses: ./
       with:
         ruby-version: ${{ matrix.ruby }}
+        bundler-cache: true
     - run: ruby -v
 
     - name: build compiler
--- a/README.md
+++ b/README.md
@@ -116,8 +116,27 @@ Otherwise, the latest Bundler version is installed (except for Ruby 2.2 and 2.3 
 
 This behavior can be customized, see [action.yml](action.yml) for details about the `bundler` input.
 
-### Caching `bundle install`
+### Caching `bundle install` automatically
 
+This action provides a way to automatically run `bundle install` and cache the result:
+```yaml
+    - uses: ruby/setup-ruby@v1
+      with:
+        bundler-cache: true
+```
+
+This caching speeds up installing gems significantly and avoids too many requests to RubyGems.org.  
+It needs a `Gemfile` under the [`working-directory`](#working-directory).  
+The caching works whether there is a `Gemfile.lock` or not.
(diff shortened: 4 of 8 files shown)


eregon:
This seems to work now.
First run: https://github.com/ruby/setup-ruby/runs/681271991?check_suite_focus=true
Cached: https://github.com/ruby/setup-ruby/runs/681303958?check_suite_focus=true

eregon:
What should happen for `-head` versions?

* For `truffleruby-head` it works fine because the commit is part of the ABI version (`RbConfig::CONFIG["ruby_version"]`).
* For `ruby-head/debug/mswin/mingw` it seems a problem as they all report ABI version `2.8.0` but I guess the ABI could change at any time. We could use `RUBY_REVISION` (the commit hash) as part of the cache key maybe? Then it could be cached if using the same build (e.g., multiple CI runs in a day, ruby-head is updated daily).
* For `jruby-head` maybe it doesn't matter since there are no native extensions?

action.yml
```diff
@@ -13,6 +13,10 @@ inputs:
     description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. The default tries Gemfile.lock and otherwise uses latest.'
     required: false
     default: 'default'
+  bundle-install:
+    description: 'Run "bundle install", and cache the result automatically.'
+    required: false
+    default: 'true'
```
eregon, line 19: I'd like to have some feedback on new inputs for this PR.
I think caching should be enabled by default because it's a major time saver (and reduces load to rubygems.org).

I don't think it's useful to have an input for "bundle install but don't cache" as users can just `- run: bundle install` for that.

I think we should try to cache both when `Gemfile.lock` exists and when it doesn't.
If there is no `Gemfile.lock`, we'll resolve the gems and install newer gems if there is any new release (both `bundle check` and `bundle install` do that if no Gemfile.lock).

dentarg:
>  test what happens for cron actions

With https://github.com/actions/cache/pull/263, using `actions/cache@master` works for `schedule` events, I've tried it on a private repo. It will be included in `actions/cache@v2`, not sure when it will be released, but it looks like it will happen soon: https://github.com/actions/cache/issues/323

When I was using `actions/cache@v1` with `schedule` events, it was a NO-OP, it printed this warning:

>[warning]Event Validation Error: The event type schedule is not supported. Only push, pull_request events are supported at this time.

index.js
```diff
@@ -177,4 +183,48 @@ async function installBundler(platform, rubyPrefix, engine, rubyVersion) {
...
+  }
+
+  let alreadyInstalled = false
+  if (cachedKey === key) {
+    const exitCode = await exec.exec('bundle', ['check'])
+    alreadyInstalled = (exitCode === 0)
+  }
+
+  if (!alreadyInstalled) {
+    await exec.exec('bundle', ['install', '--jobs', '4'])
```
dentarg, line 223: https://github.com/actions/cache/blob/9ab95382c899bf0953a0c6c1374373fc40456ffe/examples.md#ruby---bundler also have the `--retry` flag in its example (`--retry 3`)

I guess it doesn't hurt for this action to use it too?

dentarg:
> With [actions/cache#263](https://github.com/actions/cache/pull/263), using `actions/cache@master` works for `schedule` events, I've tried it on a private repo. It will be included in `actions/cache@v2`, not sure when it will be released, but it looks like it will happen soon: [actions/cache#323](https://github.com/actions/cache/issues/323)

v2 now released: https://github.com/actions/cache/releases/tag/v2.0.0

action.yml
eregon: I've renamed the input to `bundler-cache`

MSP-Greg:
The following seems like a breaking change?  Workflows that use bundler already have the code to use it, and workflows that don't...

```yaml
  bundler-cache:
    description: 'Run "bundle install", and cache the result automatically. Either true or false.'
    required: false
    default: 'true'
```

eregon:
@MSP-Greg I'm not sure, it seems harmless to `bundle install` twice, isn't it?
And `bundle install` is not attempted if `bundler: none`.

MSP-Greg:
I'm not sure either.

> it seems harmless to bundle install twice, isn't it?

I thought about whether that is true or not.  It should just result in a lot of 'using gem X' statements, but I haven't tried it with all versions, etc.  The possibility of multiple caches exists?

eregon:
I'll try with MSpec and Puma's CI at least to make it works, even with the extra `bundle install`.

> Also, some Gemfiles are largely default and bundled gems, so they don't use a bundler path?

That seems rare to me, why having a Gemfile at all then?
In such a case there is nothing to cache, but I don't think it takes long to figure that out.

I would assume most usages of `ruby/setup-ruby` use `bundle install`, so this should be a clear gain for those cases.

MSP-Greg:
> why having a Gemfile at all then?

Sometimes it's just to maintain minimum versions...

eregon:
It works fine for MSpec and for Puma, even with `bundle install` left in:
https://github.com/eregon/puma/runs/745379931?check_suite_focus=true

eregon:
@MSP-Greg @dentarg I think this PR is almost ready. Feedback welcome.

Since it's a large change, could you test it on a couple repos you know?
I'd suggest just using this PR leaving the `bundle install` in, then remove the `bundle install` and compare how much time is spent installing gems.
I did that for MSpec and Puma and they worked fine.

You can use this to test it (the full sha is ugly, but just using `@automated-bundler-caching` seems to sometimes result in older versions of the branch):
```yml
    - uses: eregon/setup-ruby@240a71d6b9d31971b61cec2a40d2f01ba703a0ad
      with:
        ruby-version: ${{ matrix.ruby }}
    - name: ragel
```

MSP-Greg:
@eregon

One thing that **might** be an issue, or at least worth mentioning in a README.  setup-ruby-pkgs installs packages.  Normally, they're needed for the repo/gem being tested (like a `rake compile` step), but, some may be needed to install dependencies.  I can't think of an example.

eregon:
True, that might be an issue. For instance `-dev/-devel` packages are needed to install `sqlite3`/`mysql2`/`pg`/etc. Some of them might be pre-installed by default in the GHA images but not all.

So maybe `setup-ruby-pkgs` should install packages before it runs `setup-ruby`? Or at least for those packages which don't depend on the Ruby version? Could you summarize which packages need to know the Ruby version, or information from running Ruby?

We could also split [this function](https://github.com/ruby/setup-ruby/blob/ba5b846810dc72703b82ed38bd4fa6d7941b3209/index.js#L26) in two so `setup-ruby-pkgs`'s package installation could run in the middle, but that might be confusing, and harder to compose the output.

MSP-Greg:
JFYI, the MSYS2 install in Windows is mostly compile tools, there are no packages that are for apps/database/etc.

At present, the only code that installs packages dependent on the Ruby version is Windows OpenSSL.  I'll look at re-arranging things.

Re running specific branches for testing, I think I'll add something (setup-ruby-loc maybe), so to pickup what here, one would use:
```
setup-ruby-loc: https://raw.githubusercontent.com/eregon/setup-ruby/automated-bundler-caching
```

Thoughts?

eregon:
> Re running specific branches for testing

Right, I'd name it `setup-ruby-ref` then since I think it can be pretty much any git `ref`.

A simpler option as long as `setup-ruby-pkgs` just calls `setup-ruby` first is:
```yml
      - uses: eregon/setup-ruby@240a71d6b9d31971b61cec2a40d2f01ba703a0ad
        with:
          ruby-version: ${{ matrix.ruby }}
      - uses: MSP-Greg/setup-ruby-pkgs@v1
        with:
          ruby-version: none
          apt-get: ragel
          ...
```

I don't intend to have many branches/big changes like this, so probably it's not really needed, but up to you.

eregon:
Ah, I missed that of course in this case the repo name also changes for `setup-ruby-loc`.
I'd say too much complexity for now, let's use the separate steps for testing.

eregon:
Another pragmatic approach would be to create a branch `automated-bundler-caching` for `setup-ruby-pkgs` (and change the URL there) just to test this PR.

MSP-Greg:
> I don't intend to have many branches/big changes like this

I think we've all thought similar things, only to...

I'm working on this...

MSP-Greg:
@eregon

I pushed the change adding the input
```yaml
setup-ruby-ref: eregon/setup-ruby/automated-bundler-caching
```

and it works fine in Puma, as you mentioned.  In my Puma fork, ran it, see the log:
https://github.com/MSP-Greg/puma/runs/745671344?check_suite_focus=true#step:3:16.  Normally (without the input) it will show `Running ruby/setup-ruby/v1`

I'll check a few more repo's...

In a little while I'll start the package install re-org...

MSP-Greg:
@eregon

Re the package problem, I gave it a bit of thought, with two things as issues.

First, as mentioned, the OpenSSL package is dependent on the Ruby version.  I forgot that other packages are also version dependent as to whether they're and old or new 'devkit' packages (MSYS vs MSYS2 or RubyInstaller vs RubyInstaller2).  Secondly, setup-ruby-pkgs needs to know whether a Windows job is mingw or mswin.

I came up with a few ways of doing it, and finally decided that the best was to pass a function to the run function.  See index.js in the commit in my fork: https://github.com/MSP-Greg/ruby-setup-ruby/commit/909ed251c83

I ran it both by itself and within setup-ruby-pkgs modified for it, and both ran fine.  Thoughts?

eregon:
If the existing workflow already uses `bundle install --path some/path` then it will actually install gems twice which is suboptimal, but still works:
https://github.com/eregon/puma/runs/746891598?check_suite_focus=true

Another idea for more flexibility would be to have a `ruby/bundle-install` action or so, and deal with bundler there, not in this action. I'm not sure there are many use cases for that though, except the packages installation discussed here.

The idea to pass a function seems nice.
I could also call that function with `(platform, engine, version, rubyPrefix)` which might be more convenient.
The callback reverses the what wraps what in the output, but we can probably make it look OK (i.e., setup-ruby output will wrap packages installation output, while setup-ruby-pkgs as an action wraps setup-ruby).
Does it need to run after `gem install bundler`, or could it run before, here:
https://github.com/ruby/setup-ruby/blob/ba5b846810dc72703b82ed38bd4fa6d7941b3209/index.js#L49 ?

MSP-Greg:
@eregon

> Does it need to run after gem install bundler, or could it run before

Asking myself why I put it there.  Can't see any reason for putting it inside the conditional, so I'll move above.

> I could also call that function with (platform, engine, version, rubyPrefix) which might be more convenient.

I wondered about that.  Assuming that they're passed by reference, and given that there is code here running after the call, maybe best to not pass parameters?

> If the existing workflow already uses bundle install --path some/path then it will actually install gems twice which is suboptimal

Another example is ruby/reline, where the Gemfile contains nothing but gems that are wither default gems or bundled gems, and hence, quite often uses the existing installed gems.

Both of which get back to the idea that this should be an 'opt-in', not an 'opt-out'?

Re the console output, we can fix that easily...

MSP-Greg:
I cleaned up the logging in pkgs, see:
https://github.com/MSP-Greg/setup-ruby-pkgs/runs/747430419?check_suite_focus=true#step:3:10

Obviously, in a normal run, `ruby/setup-ruby/v1` would replace `MSP-Greg/ruby-setup-ruby/bundler-cache`...

dentarg:
> Since it's a large change, could you test it on a couple repos you know?

I've tested it here now: https://github.com/Starkast/wikimum/actions?query=branch%3Aactions%2Fsetup-ruby-cache

* https://github.com/Starkast/wikimum/runs/750749877?check_suite_focus=true
* https://github.com/Starkast/wikimum/runs/750804413?check_suite_focus=true

Also tested it on a private repo, works there too

I usually use [vendoring](https://bundler.io/v2.0/man/bundle-package.1.html) in projects, and it looks like the first `bundle install` is slower when it is not using `--local`, see the first run above (81.83 seconds) compared to this run: https://github.com/Starkast/wikimum/runs/744631746?check_suite_focus=true#step:8:1 (71 seconds). Maybe not scientific enough :)

eregon:
I'm going to merge, disabled by default for now, so it should be safe and we can experiment easily with it with `bundler-cache: true`.

PR #59, opened 2020-06-09
Update README.md

imbsky:


imbsky:
Yeah, I was aware of that. But I'm not a big fan of it.

eregon:
@imbsky Could you expand a bit more on why you don't like it?

imbsky:
It’s likely to be just a matter of preference. I just don't like adding such options to make the action fat. Also, if there are few dependencies, it can be slow, and if there are many, it will reach the cache limit and thrashing will occur. And in that case, it's going to be incredibly slow. At this point, the cache feature needs to be used at your discretion by people who know its advantages and disadvantages.

imbsky:
To be clear, I’m a dedicated Actions contributor and one of the people who recently improved cache action. In spite of my position, I don’t yet want to recommend it.

imbsky:
I intended this.
https://en.wikipedia.org/wiki/Thrashing_(computer_science)
Well, I don't how bundler cache works so I may have missed something. But in my experience, 5GB is not enough. In many cases, users will just pay extra costs for download/upload and decompression/compression times.

imbsky:
But it seems pretty good that it's disabled by default. I'm still not a fan, but I really like it. and if 5GB isn't enough, GitHub will consider bumping the limit.

eregon:
Thrashing seems related to swapping virtual memory to disk. I don't think the `cache` action would cause swapping. I would assume the compression/decompression is done incrementally to avoid high memory usage.

dentarg:
@eregon I think ["cache thrashing"](https://en.wikipedia.org/wiki/Thrashing_(computer_science)#Other_uses) could be a real concern _if_ you have _many_ gems. I think the concerns are that you would pay time for compressing and uploading content for a cache that wont really be used.

I don't work with any Ruby projects that large were this would be a real concern, so I can't really judge how common of a problem it would be. I think opt-ing out of the automated caching would be an easy fix for those who run into the problem.

imbsky:
Yes, @dentarg’s comment is correct.

> I think opt-ing out of the automated caching would be an easy fix for those who run into the problem.

Yeah, probably.

PR #60, opened 2020-06-09
Test all virtual environments, not just the latest, and added crontab 

jwillemsen:


jwillemsen:
Looks on windows-2016 jruby-9.1/jruby/jruby-head/mswin are all failing, are these known failures which I should exclude?

jwillemsen:
jruby seems to fail because tar doesn't seem to be on the expected place on windows-2016 and mswin assumes Visual Studio 2019 which is not there on windows-2016

MSP-Greg:
Thanks.

The problem is that a native tar.exe exists in Windows 2019, but not Windows 2016.  Maybe you can experiment in your fork, as the Git or MSYS2 version may work.  I can look at it later...

.github/workflows/test.yml
```diff
@@ -8,25 +8,34 @@ on:
...
     - README.md
+  schedule:
+    - cron: '0 7 * * SUN'
 jobs:
   test:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu, macos, windows ]
+        os: [ ubuntu-16.04, ubuntu-18.04, ubuntu-20.04, macos-10.15, windows-2016, windows-2019 ]
```
eregon, line 18: Could you remove `windows-2016` here since it's not supported (see #61 and the README)?
And then many excludes are no longer needed.

jwillemsen: See https://github.com/actions/virtual-environments#available-environments, there it is still listed as available environment, not sure which docu is now correct, but windows-2016 does seem to work

eregon:
FWIW It was intentional to run a subset of platforms so the CI is reasonably fast and we don't create so many jobs. 74 jobs is still probably somewhat reasonable (but will result in a lot of queuing), so I'm OK to try with this PR and see how it goes.

Note that testing all Ruby versions on all platforms would result in hundreds of jobs and so is completely unrealistic. I don't think it's needed nor very useful either, so the proposed matrix in this PR seems OK.

.github/workflows/test.yml
```diff
@@ -8,25 +8,34 @@ on:
...
         exclude:
-        - { os: windows, ruby: debug }
-        - { os: windows, ruby: truffleruby }
-        - { os: windows, ruby: truffleruby-head }
+        - { os: windows-2016, ruby: debug }
+        - { os: windows-2016, ruby: truffleruby }
+        - { os: windows-2016, ruby: truffleruby-head }
+        - { os: windows-2016, ruby: jruby }
+        - { os: windows-2016, ruby: jruby-9.1 }
+        - { os: windows-2016, ruby: jruby-head }
```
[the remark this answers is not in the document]
jwillemsen: done

PR #62, opened 2020-06-09
Fix tar extraction with windows-2016 image

MSP-Greg:
1. common.js - add function win2nix(path) to convert windows style paths to nix style

2. ruby_builder.js - use Git tar for windows-2016, leave windows-2019 as is

3. Add three windows-2016 JRuby builds to test.yml


MSP-Greg:
We can add more tests, but the extraction was the main issue.

I left the logic the same as before for windows-2019, so when windows-2016 is removed, the logic using Git tar can be removed.

The mswin build does not work with windows-2016, as extension gems will not compile with VS 2017 and the build's libraries.  I really don't see a need for generating another version.

Issue #57, opened 2020-05-21
setup-ruby interferes with Windows runner's own MSYS2 install


lopopolo:
https://github.com/ruby/setup-ruby/blob/002b15eeca2c8a043d93bfd8bd5acd44761954f9/windows.js#L121-L124

https://github.com/actions/virtual-environments/issues/905#issuecomment-631741076

I believe it is time to remove this bit of setup. If it is not, how can I opt out of setup-ruby including a bundled MSYS2?

MSP-Greg:
@lopopolo

See my response in https://github.com/actions/virtual-environments/issues/905.

Re changes here, there are quite a few repos running Windows CI, and having MSYS2 in the path is essentially expected.  An input could be added to allow it not to be so, but I think the better fix is to remove 7z from the MSYS2 install...

eregon:
That specific case seems a bug of upstream:
https://github.com/actions/virtual-environments/issues/905#issuecomment-631872617
https://github.com/actions/virtual-environments/pull/916

The linked code will actually do nothing now, right? Since C:\msys64 will exist (but it's likely not rolled out everywhere yet so we need the fallback).

I guess the only to do is we shouldn't prepend `C:\msys64\mingw64\bin` and `C:\msys64\usr\bin` to the PATH if `C:\msys64` already exists, and if MSYS2 is already in the PATH.
That way the PATH is closer to what it looks like when the runner starts.
But from what I heard MSYS2 isn't added to PATH by default, so I guess no changes at all?

eregon:
I tried to make sure my assumptions regarding PATH are correct in https://github.com/ruby/setup-ruby/runs/695926638?check_suite_focus=true
But that revealed C;\msys64 doesn't exist in that run and so is not yet fully deployed.

eregon:
When there is no C:\msys64 in the image, the 7z from Chocolatey is used (both before and after this action):
https://github.com/ruby/setup-ruby/runs/695968983?check_suite_focus=true

MSP-Greg:
> I guess no changes at all?

Agreed.

MSP-Greg:
JFYI, the comment below is incorrect.

The code should be removed when Actions removes the embedded MSYS2 installs that currently exist in the four ToolCache Rubies.

When that happens (only after MSYS2 install is rolled out), there will be nothing to symlink to.

```
// Remove when Actions Windows image contains MSYS2 install
```

PR #64, opened 2020-06-12
ENV setup, Windows tar, Windows bash shell

### LINKED ISSUE #81: sassc-ruby fails to install on Windows + JRuby
See [test project](https://github.com/slonopotamus/sassc-jruby-test/blob/master/Gemfile) and [build log](https://github.com/slonopotamus/sassc-jruby-test/runs/1026295911?check_suite_focus=true#step:4:17).

I've also reported this bug as sass/sassc-ruby#214 because I'm not sure what party is to blame here.

MSP-Greg:
Three Commits:

A. ENV setup, Windows tar, Windows bash shell

B. Add envPreInstall, common.setupPath, use MSYS2 bash for Windows bash

1. index.js - envPreInstall - sets ENV values for runners

2. common.js - setupPath - collects all Path operations into one function, runs before Ruby install

3. Add MSYS2 paths to all Windows builds for MSYS2 Bash use

C. Add Actions 'bash test' and 'Windows JRuby' (gem install sassc) steps

Closes #81

### DIFF
--- a/common.js
+++ b/common.js
@@ -1,10 +1,13 @@
 const os = require('os')
+const path = require('path')
 const fs = require('fs')
 const util = require('util')
 const stream = require('stream')
 const crypto = require('crypto')
 const core = require('@actions/core')
 const { performance } = require('perf_hooks')
+
+const isWin = (os.platform() === 'win32')
 
 export async function measure(name, block) {
   return await core.group(name, async () => {
@@ -61,4 +64,31 @@ export function win2nix(path) {
     path = `/${path[0].toLowerCase()}${path.split(':', 2)[1]}`
   }
   return path.replace(/\\/g, '/').replace(/ /g, '\\ ')
+}
+
+export function setupPath(newPathEntries) {
+  const envPath = isWin ? 'Path' : 'PATH'
+  const originalPath = process.env[envPath].split(path.delimiter)
+  let cleanPath = originalPath.filter(entry => !/\bruby\b/i.test(entry))
+
+  if (cleanPath.length !== originalPath.length) {
+    core.startGroup(`Cleaning ${envPath}`)
+    console.log(`Entries removed from ${envPath} to avoid conflicts with Ruby:`)
+    for (const entry of originalPath) {
+      if (!cleanPath.includes(entry)) {
+        console.log(`  ${entry}`)
+      }
+    }
+    core.exportVariable(envPath, cleanPath.join(path.delimiter))
+    core.endGroup()
+  }
+  let newPath
+  if (isWin) {
+    // add MSYS2 path to all for bash shell
+    const msys2 = ['C:\\msys64\\mingw64\\bin', 'C:\\msys64\\usr\\bin']
+    newPath = [...newPathEntries, ...msys2]
+  } else {
+    newPath = newPathEntries
+  }
+  core.addPath(newPath.join(path.delimiter))
 }
--- a/index.js
+++ b/index.js
@@ -6,6 +6,8 @@ const exec = require('@actions/exec')
 const cache = require('@actions/cache')
 const common = require('./common')
 
+const isWin = (os.platform() === 'win32')
+
 const inputDefaults = {
   'ruby-version': 'default',
   'bundler': 'default',
@@ -26,7 +28,7 @@ export async function run() {
 export async function setupRuby(options = {}) {
   const inputs = { ...options }
   for (const key in inputDefaults) {
-    if (!inputs.hasOwnProperty(key)) {
+    if (!Object.prototype.hasOwnProperty.call(inputs, key)) {
       inputs[key] = core.getInput(key) || inputDefaults[key]
     }
   }
@@ -48,9 +50,10 @@ export async function setupRuby(options = {}) {
 
   createGemRC()
 
-  const [rubyPrefix, newPathEntries] = await installer.install(platform, engine, version)
+  envPreInstall()
 
-  setupPath(newPathEntries)
+  const rubyPrefix = await installer.install(platform, engine, version)
+
 
   // When setup-ruby is used by other actions, this allows code in them to run
   // before 'bundle install'.  Installed dependencies may require additional
@@ -135,22 +138,18 @@ function createGemRC() {
   }
 }
 
-function setupPath(newPathEntries) {
-  const originalPath = process.env['PATH'].split(path.delimiter)
-  let cleanPath = originalPath.filter(entry => !/\bruby\b/i.test(entry))
-
-  if (cleanPath.length !== originalPath.length) {
-    core.startGroup('Cleaning PATH')
-    console.log('Entries removed from PATH to avoid conflicts with Ruby:')
-    for (const entry of originalPath) {
-      if (!cleanPath.includes(entry)) {
-        console.log(`  ${entry}`)
-      }
-    }
-    core.endGroup()
+// sets up ENV variables
+// currrently only used on Windows runners
+function envPreInstall() {
+  const ENV = process.env
+  if (isWin) {
+    // puts normal Ruby temp folder on SSD
+    core.exportVariable('TMPDIR', ENV['RUNNER_TEMP'])
+    // bash - sets home to match native windows, normally C:\Users\<user name>
+    core.exportVariable('HOME', ENV['HOMEDRIVE'] + ENV['HOMEPATH'])
+    // bash - needed to maintain Path from Windows
+    core.exportVariable('MSYS2_PATH_TYPE', 'inherit')
   }
-
-  core.exportVariable('PATH', [...newPathEntries, ...cleanPath].join(path.delimiter))
 }
 
 function readBundledWithFromGemfileLock(path) {
--- a/ruby-builder.js
+++ b/ruby-builder.js
@@ -9,23 +9,29 @@ const rubyBuilderVersions = require('./ruby-builder-versions')
 const builderReleaseTag = 'enable-shared'
 const releasesURL = 'https://github.com/ruby/ruby-builder/releases'
 
+const isWin = (os.platform() === 'win32')
+
 export function getAvailableVersions(platform, engine) {
   return rubyBuilderVersions.getVersions(platform)[engine]
 }
 
 export async function install(platform, engine, version) {
   const rubyPrefix = await downloadAndExtract(platform, engine, version)
-  let newPathEntries
-  if (engine === 'rubinius') {
-    newPathEntries = [path.join(rubyPrefix, 'bin'), path.join(rubyPrefix, 'gems', 'bin')]
-  } else {
-    newPathEntries = [path.join(rubyPrefix, 'bin')]
-  }
-  return [rubyPrefix, newPathEntries]
+  return rubyPrefix
 }
 
 async function downloadAndExtract(platform, engine, version) {
-  const rubiesDir = path.join(os.homedir(), '.rubies')
+  const rubiesDir = isWin ?
+    `${(process.env.GITHUB_WORKSPACE || 'C')[0]}:` :
+    path.join(os.homedir(), '.rubies')
+
+  const rubyPrefix = path.join(rubiesDir, `${engine}-${version}`)
+  const newPathEntries = (engine === 'rubinius') ?
+    [path.join(rubyPrefix, 'bin'), path.join(rubyPrefix, 'gems', 'bin')] :
+    [path.join(rubyPrefix, 'bin')]
+
+  common.setupPath(newPathEntries)
+
   await io.mkdirP(rubiesDir)
 
   const downloadPath = await common.measure('Downloading Ruby', async () => {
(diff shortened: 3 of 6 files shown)


MSP-Greg:
Re CI:

1. Maybe split it into a few workflows by OS?  The number of jobs sort of breaks the web UI.  Also, remove a few jobs?  The full duplication when running different OS versions seems overkill.

2. Wondering if a Ruby test script may be helpful for CI, at least on Windows.  Things like whether `Gem.path` contains two items, `Gem env` has the proper shell entries, etc

`Path` manipulation is really odd, almost unstable. As mentioned, the 'commands' in core run thru stdout, and it may be async depending on the io type and OS.  See https://nodejs.org/api/process.html#process_a_note_on_process_i_o.

I need to look into this more.  Since the stdout syntax is published, I may look at using a stdout.write to combine several Path and/or env variable statements.  I believe the core methods both modify ENV variables in the node environment and write to stdout for the C# runner to process the command...

MSP-Greg:
Updated to write 'core commands' directly to the runner.  This fixes the step env display, and also eliminates rare intermittent failures.

eregon:
> Updated to write 'core commands' directly to the runner. This fixes the step env display, and also eliminates rare intermittent failures.

Could you rather fix it upstream in https://github.com/actions/toolkit?
I'd much prefer to use official APIs than hardcoding it ourselves.

MSP-Greg:
I believe it's an official API, see
https://help.github.com/en/actions/reference/workflow-commands-for-github-actions

All that actions/core does is combine the above with doing the same in the current node process.

eregon:
Yes I know, but wouldn't it be good to fix it (i.e., make it a synchronous write) for all actions, not just this one?
Manually writing those still feels suboptimal to me, especially when there is a proper API designed for it.

MSP-Greg:
In theory, it is a sync write on Windows (`Boolean(process.stdout.isTTY)` is `false`), but as mentioned, I've seen intermittent errors.  Also, behavior varies by OS.  And that's just on the node/js side, C# is another issue.

At present, we're using very few ENV calls for Ubuntu or macOS, but that could change.

> there is a proper API designed for it

actions/core is a 'helper' action.  It may have not been tested against all possible uses.  The API is the pipe text description, actions/core is a wrapper around that.

JFYI, js `split(':', 2)` is equal to Ruby `split(':')[0, 2]`, not Ruby `split(':', 2)`.  Either way, `:` is not a valid  directory or file name character in Windows.

MSP-Greg:
Added a class to common.js that handles ENV operations.  Replaces all calls to `core.exportVariable` & `core.addPath`.  As mentioned, the problem this fixes is intermittent, causing failures in less that 5% of jobs (may actually be more like < 1%).

common.js
```diff
@@ -51,3 +52,50 @@ export function win2nix(path) {
...
+
+class CmdCls {
+  constructor() {
+    this.varis = []
+    this.paths = []
+    this.os_path = os.platform() === 'win32' ? 'Path' : 'PATH'
+    this.ENV = process.env
+  }
+  
+  addVari(k,v) {
```
eregon, line 64: I'd like to keep this action source code, so no abbreviations please.

eregon:
> Added a class to common.js that handles ENV operations. Replaces all calls to `core.exportVariable` & `core.addPath`. As mentioned, the problem this fixes is intermittent, causing failures in less that 5% of jobs (may actually be more like < 1%).

Can you file a bug to that repo then? And do you have a log of a case that failed.
Maybe it failed for another reason than we expect, or a bug in the runner, etc.
i.e, I'm not convinced `core.addPath/exportVariable` are to blame here, and if they are we should fix those, not duplicate the logic here.

eregon:
> The number of jobs sort of breaks the web UI.

What do you mean? It seems to work fine for me (except being a long list).

eregon:
> 2\. Wondering if a Ruby test script may be helpful for CI, at least on Windows.

Yes, I think that would be useful, and avoid so much commands in test.yml.

MSP-Greg:
> What do you mean? It seems to work fine for me (except being a long list).

If one is using one tab and trying to look at Windows logs, one is constantly scrolling the window.  PITA.  Bad UI, the primary pane (the job list) should sroll indenpendently of the secondary, dependent pane, the log.

> Yes, I think that would be useful, and avoid so much commands in test.yml.

Some I think should still be ran, but they can run within the script.  It's helpful to chceck if the Windows Rubies have working bash scripts for all the bin files?

> I don't want to duplicate code...

I'm working on it.  The issue may just be the handling of 'Path', so I'm thinking the common.cmd class maybe should handle core.addPath, along with `CleanPath`, and use `core.exportVariable` as before...

MSP-Greg:
@eregon

I got it working using nothing but the core functions.  PITA.  Most of it seems to be due to removing items from `Path`, which I think should be done when possible.  The most important thing is cleaning the `Path` first, then having all the core.addPath statements after that.

One odd thing is that the `Path` shown in the first 'ENV' group in a step (see https://github.com/MSP-Greg/ruby-setup-ruby/runs/771043663?check_suite_focus=true#step:17:11) is the path after cleaning, but it doedn't show any additions using `core.addPath`.  See the path output from `gem env` later on in the same step.

Anyway, all Windows Rubies are running in a bash shell, using their respective devkits.

If PR #65 is okay, I'll revise this after that's merged?

eregon:
#65 is merged now, I am sorry about the delay.
Being able to just use the `core` functions sounds great.

Regarding removing the system ruby from PATH I'm not sure if it's very important. Maybe we could drop that and only use addPath()? Indeed, the output the show seems like some kind of bug in actions, I think it would be worth reporting to them.

The Ruby we setup should anyway have all executables the system ruby has + some extra, no? And on many development systems it's common to have a system ruby + RVM/rbenv/chruby and they don't remove anything from PATH, so it seems not so much an issue, and something Rubyists already have to deal with.
Is there anything Windows-specific regarding that?
I think maybe JRuby on Windows was confusing as it could sometimes pick the system Ruby's binaries when using `ruby ...` or `ruby.exe ...`.
So if no longer cleaning the PATH simplifies things or makes the action more reliable I think we could no longer remove the "system Ruby" from PATH.

MSP-Greg:
> I am sorry about the delay.

No problem, as it only affects the 'automatic' bundle iinstall.  I mentioned revising this, and it's been a couple of weeks.  I can't recall.  I believe it was stable with the removal.

eregon:
It would also fix using the correct Bash on Windows, right (MSYS2' and not Git' Bash)?
That seems an important fix, to avoid issues like #57.

MSP-Greg:
@eregon

Well, it's mostly finished, see https://github.com/MSP-Greg/ruby-setup-ruby/runs/837069369.

I currently have some Bash/PowerShell code contained in it for CI.  Since that code is really about the Windows Ruby builds, there isn't a need in the CI here, at least not on an 'every commit' basis.  The bash rake binstub in 2.3 has a bad shebang line, and there maybe something missing in the ruby-head build (bundler?).  The issue is that the Windows Ruby's bin folders need files for both cmd/bat usage, and bash usage, and the bash files haven't been used by a lot of people...

Re Path issues, I changed the code back to a single `core.exportVariable('PATH', ...)`, and it's displaying fine in the Actions UI.  Path of the Path issue was related to needing a tar command, as Windows 2016 doesn't have a native one like Windows 2019/10.  I changed the code to use Git's tar...  I'll revise it and open a PR sometime soon.

eregon:
https://github.com/MSP-Greg/ruby-setup-ruby/runs/837069345 seems to run in the Git Bash shell (and not the MSYS2 shell). Or is that just a UI bug?

MSP-Greg:
I hate this.  Apologies, you're correct, it is the Git shell.

You will note that the log is showing PATH as is set in the code, but it's using the wrong shell.  Conversely, in the older code, the Path was shown incorrectly, but the MSYS2 bash shell was used.

I may try to work the casing of Path.  The link above shows 'PATH', but as shown at https://github.com/MSP-Greg/ruby-setup-ruby/runs/777119210?check_suite_focus=true#step:4:7, the Path is incorrect, but it is shown as 'Path', and the MSYS2 is used.

So the issue is either the casing, in which case we may be able to get the path shown correctly and the right shell, or the issue is that the C# runner only recognizes path changes via `addPath`, in which case we probably have to live with an incorrect log of Path.  I think that made sense.

Sorry for mixing things, but Ruby ABI is a contract/promise.  What we're doing here seems to be a grey area where the Actions API doesn't really promise anything.  Let me see if the casing has any affect...

MSP-Greg:
The only way I could get it working causes the logged path to be incorrect.  See:
https://github.com/MSP-Greg/ruby-setup-ruby/actions/runs/158427922

One change is that the MSYS2 paths are added to every Windows build, allowing JRuby CI to run with Bash.  For older RI/MSYS builds (2.2 & 2.3), the MSYS paths are added before the MSYS2 paths.

I need to dice & chop it into more commits...

eregon:
> One change is that the MSYS2 paths are added to every Windows build, allowing JRuby CI to run with Bash

Sounds good.

Would using only `addPath` and not `exportVariable("PATH", value)` work both in the UI and use the MSYS shell?
That sounds kind of tempting.

Probably we should report a bug to https://github.com/actions/runner that `exportVariable("PATH", value)` doesn't work well.

MSP-Greg:
> Would using only addPath and not exportVariable("PATH", value) work both in the UI and use the MSYS shell?
That sounds kind of tempting.

Actually, I think I mixed things up.  `addPath` does not show in the UI, only variables that have changes via `exportVariable`.  Since both are being used, I got a bit confused.

I haven't looked at the code for a bit, I did change a few things, mostly to collect all the PATH manipulation into one function, so if the API is changed/updated, it's all in one place.

See https://github.com/actions/runner/issues/576, I just added a comment about adding a function like `removeTool`, which is all we'd need.  I think.

MSP-Greg:
Updated.  All Windows Rubies now have MSYS2 added to path, so the JRuby issue is fixed, and the MSYS2 bash shell is the Windows default.

slonopotamus:
> and the MSYS2 bash shell is the Windows default.

Just to make it clear: this PR will change PowerShell to Bash for any workflow that uses `ruby/setup-ruby` or I am misunderstanding something?

MSP-Greg:
Sorry I wasn't a bit more clear.

The default shell in Windows should remain as PowerShell, as changing that would break a lot of CI scripts/workflows.

If you run an Action step in a bash shell by changing the default or with the `shell: bash` setting, the MSYS2 bash shell will be used, not the Git bash shell.

The Git bash shell is an MSYS2 bash shell, but may have a few less commands than the MSYS2 shell currently installed in the Actions Windows runners.  The Git shell is probably missing some of the more dev oriented commands.

See:
https://github.com/ruby/setup-ruby/runs/1029689743#step:16:3

vs the current behavior:
https://github.com/rubygems/rubygems/runs/1036211465?check_suite_focus=true#step:5:3

slonopotamus:
> If you run an Action step in a bash shell by changing the default or with the shell: bash setting, the MSYS2 bash shell will be used, not the Git bash shell.

Understood, thanks.

MSP-Greg:
Since this PR changes the bash shell, and Bundler's Windows CI is ran from a bash shell, I ran its CI using this PR.

All CI passed.

See:
https://github.com/MSP-Greg/rubygems/actions/runs/227189348

eregon:
This looks great, I'll try to review soon.

index.js
```diff
@@ -26,7 +28,7 @@ export async function run() {
 export async function setupRuby(options = {}) {
   const inputs = { ...options }
   for (const key in inputDefaults) {
-    if (!inputs.hasOwnProperty(key)) {
+    if (!Object.prototype.hasOwnProperty.call(inputs, key)) {
```
eregon, line 31: Is this needed? What does it do differently?

common.js
```diff
@@ -62,3 +65,30 @@ export function win2nix(path) {
...
+  }
+  let newPath
+  if (isWin) {
+    // add MSYS2 path to all for bash shell
+    const msys2 = ['C:\\msys64\\mingw64\\bin', 'C:\\msys64\\usr\\bin']
+    newPath = [...newPathEntries, ...msys2]
+  } else {
+    newPath = newPathEntries
+  }
+  core.addPath(newPath.join(path.delimiter))
```
eregon, line 93: A loop over paths with a ` core.addPath()` seems nicer (but would need to be done in reverse order).
Since this work though it seems fine.

ruby-builder.js
```diff
@@ -9,23 +9,29 @@ const rubyBuilderVersions = require('./ruby-builder-versions')
...
+  const rubiesDir = isWin ?
+    `${(process.env.GITHUB_WORKSPACE || 'C')[0]}:` :
+    path.join(os.homedir(), '.rubies')
+
+  const rubyPrefix = path.join(rubiesDir, `${engine}-${version}`)
+  const newPathEntries = (engine === 'rubinius') ?
+    [path.join(rubyPrefix, 'bin'), path.join(rubyPrefix, 'gems', 'bin')] :
+    [path.join(rubyPrefix, 'bin')]
+
+  common.setupPath(newPathEntries)
```
eregon, line 33: This seems done rather early, even before the `rubyPrefix` exists. I guess we can move it at the end of this function?

windows.js
```diff
@@ -149,8 +122,9 @@ export function addVCVARSEnv() {
   let newPathEntries = undefined
   for (let [k, v] of newEnv) {
     if (process.env[k] !== v) {
-      if (k === 'Path') {
-        newPathEntries = v.replace(process.env['Path'], '').split(path.delimiter)
+      if (/^Path$/i.test(k)) {
+        const newPathStr = v.replace(`$(path.delimiter)${process.env['Path']}`, '')
```
eregon, line 126: `$(path.delimiter)` => `${path.delimiter}` right?

eregon (approved):

eregon:
@MSP-Greg Great work, thanks! I'll make a PR with small cleanups.

ruby-builder.js
eregon: Ah but for Windows it's useful to add MSYS2 in Path for extraction.

windows.js
eregon: https://github.com/ruby/setup-ruby/pull/83

eregon: I checked, that Path was doubled with that, before:
https://github.com/ruby/setup-ruby/runs/1066490911?check_suite_focus=true
(e.g., `mysql` is twice in Path)
after:
https://github.com/ruby/setup-ruby/pull/83/checks?check_run_id=1066493565

index.js
MSP-Greg: This was an ESLint thing.  I still use it locally to (hopefully) catch typo's etc.  I need a new keyboard, as my old one required a lot less travel.

ruby-builder.js
MSP-Greg: > This seems done rather early

I can't recall the timeline of this.  It involves the `afterSetupPathHook`, and the issue of installing packages before `bundle install`.  I don't think I've modified `setup-ruby-pkgs` for that, but I'm not sure.  I'm looking at it shortly.

eregon: I kept it there, it seems useful notably to get `tar` from MSYS2 on Windows.

MSP-Greg: JFYI, as I recall, that was needed only because the Win2016 image didn't have a native tar, which Win2019 did.  For instance, locally (Windows 10), I have a native Windows `tar`

PR #65, opened 2020-06-14
Allow running external function before 'bundle install'

MSP-Greg:
External actions may use setup-ruby, and some actions make changes/additions to the build environment.  These may be needed for dependencies.

Hence, allow those external actions to run code before 'bundle install'.  This is done by passing a function as a parameter to run().

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,14 +16,14 @@ const inputDefaults = {
 // entry point when this action is run on its own
 export async function run() {
   try {
-    await setupRuby({})
+    await setupRuby()
   } catch (error) {
     core.setFailed(error.message)
   }
 }
 
 // entry point when this action is run from other actions
-export async function setupRuby(options) {
+export async function setupRuby(options = {}) {
   const inputs = { ...options }
   for (const key in inputDefaults) {
     if (!inputs.hasOwnProperty(key)) {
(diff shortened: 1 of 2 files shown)


eregon:
I pushed c8a88df1573f0fb5ee0390a60688742fd9846913, could you rebase on top of that?
Then only `setupRuby` needs to be modified, not `run()`.

index.js
```diff
@@ -13,12 +13,13 @@ const inputDefaults = {
...
 }
 
-export async function run() {
+export async function run(func = null) {
   try {
     const options = {}
     for (const key in inputDefaults) {
       options[key] = core.getInput(key)
     }
+    options['func'] = func
```
eregon, line 22: After rebase, I think the changes above can be dropped

index.js
```diff
@@ -49,6 +50,13 @@ export async function setupRuby(options) {
 
   setupPath(newPathEntries)
 
+  // When setup-ruby is used by other actions, this allows code in them to run
+  // before 'bundle install'.  Installed dependencies may require additional
+  // libraries & headers, build tools, etc.
+  if (inputs['func'] instanceof Function) {
```
eregon, line 56: `func` is rather unspecific, maybe `afterSetupPathHook` ?

index.js
```diff
@@ -49,6 +50,13 @@ export async function setupRuby(options) {
...
+    await inputs['func']()
```
eregon, line 57: Should we pass it `{ rubyPrefix, platform, engine, version }`?
Then the callback might use it or not, but it's probably useful at least in some cases.

MSP-Greg: Changed the order to match a code line a few statements later.  Done

index.js
MSP-Greg: Done

index.js
MSP-Greg: Done

eregon:
I am sorry, I forgot about this PR, it looks great, I'll merge it now.

PR #66, opened 2020-06-22
Compensate for premature close errors on cache restore

elizabrock:
Our buils is running into quite a few intermittent premature stream close errors (`##[error] Premature close`) while using ruby/setup-ruby (perhaps once in every 30 runs).  I tracked the source of the error down to the `restoreCache` call, and added a retry step on restoring the cache.  I believe that a single retry will be enough in most cases because the issue is quite intermittent. 

An example:
Screen Shot 2020-06-22 at 11 06 11 AM

I believe that our build may be more apt to experience these errors because our cached bundle is rather large (~304 MB).

An example of the retry working:
Screen Shot 2020-06-22 at 11 08 01 AM

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -2420,7 +2420,19 @@ async function bundleInstall(platform, engine, version) {
   console.log(`Cache key: ${key}`)
 
   // restore cache & install
-  const cachedKey = await cache.restoreCache(paths, key, restoreKeys)
+  let cachedKey = null
+  try {
+    cachedKey = await cache.restoreCache(paths, key, restoreKeys)
+  } catch (error) {
+    var isPrematureClose = error.code === 'ERR_STREAM_PREMATURE_CLOSE'
+    if(isPrematureClose){
+      console.log("There was a premature stream close error while attempting to restore the cache. Making an additional attempt now.")
+      cachedKey = await cache.restoreCache(paths, key, restoreKeys)
+    } else {
+      throw error
+    }
+  }
+
   if (cachedKey) {
     console.log(`Found cache for key: ${cachedKey}`)
   }
--- a/index.js
+++ b/index.js
@@ -234,7 +234,19 @@ async function bundleInstall(platform, engine, version) {
   console.log(`Cache key: ${key}`)
 
   // restore cache & install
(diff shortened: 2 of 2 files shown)


eregon:
Thanks for the PR.

Do you know why does this error happen, i.e., what causes it?

It seems actions/cache doesn't retry for that case:
https://github.com/actions/cache/blob/eed9cfe64d00dd64cbb36bec915d41dd6fab9f6c/src/restore.ts#L28

I wonder if this should be fixed in https://github.com/actions/toolkit/tree/master/packages/cache instead.

elizabrock:
I believe that the underlying issue could be broadly described as "network hiccups".

It looks to me as if actions/cache chooses to treat it as a cache miss, rather than error out the build entirely: https://github.com/actions/cache/blob/eed9cfe64d00dd64cbb36bec915d41dd6fab9f6c/src/restore.ts#L55

I considered doing that instead of a retry, but it seemed more expensive (in build time) than just trying again.

That being said, I had the bad luck to get two premature close errors in a row just now, so perhaps I should adjust my code to try twice and then fall back to treating it as a cache miss.

elizabrock:
For science, I switched to using actions/cache for caching and ruby/setup-ruby for installing ruby.  This verified for me that they treat it as a warning rather than an error:

Screen Shot 2020-06-23 at 10 25 33 AM

@eregon What do you think of me making a PR that treats the premature close in the same way that actions/cache does (i.e. as a warning rather than an error)? I think that it would look something like this: https://github.com/ruby/setup-ruby/compare/master...elizabrock:treat-error-as-cache-miss

eregon:
Thanks for the investigation!
I agree, the warning sounds good, and better than a retry (which might slow things down and not reveal the source issue).

eregon:
@elizabrock Could you open a PR with that change?

eregon:
Seems I can create the PR myself with your commits :smiley: 
https://github.com/ruby/setup-ruby/pull/68

eregon:
Fixed with #68 and released as https://github.com/ruby/setup-ruby/releases/tag/v1.39.1

PR #74, opened 2020-08-11
Support multiple gemfiles

### LINKED ISSUE #73: Please consider BUNDLE_GEMFILE env variable
The current version (5a76bd6a5fe27d6e6e8af1f2e605f46d17557049) of the readme says

> By default, if there is a `Gemfile.lock` file with a `BUNDLED WITH` section,
> the latest version of Bundler with the same major version will be installed.
> Otherwise, the latest Bundler version is installed (except for Ruby 2.2 and 2.3 where only Bundler 1 is supported).
>
> This behavior can be customized, see action.yml for details about the `bundler` input.

When doing gem development it’s not uncommon to have CI testing various combinations of gem dependencies, for example 
 testing a Rails-specific library against e.g. Rails v4.2, v5.2 and v6.0.

To accomplish this, I usually have a `gemfiles` folder containing a number of gemfiles, and I would configure GitHub Actions like this:

```yaml
strategy:
  matrix:
    gemfile:
      - gemfiles/rails4.2.gemfile
      - gemfiles/rails5.2.gemfile
      - gemfiles/rails6.0.gemfile
steps:
  - uses: actions/checkout@v2
  - uses: ruby/setup-ruby@v1
  - run: echo "::set-env name=BUNDLE_GEMFILE::$GITHUB_WORKSPACE/${{ matrix.gemfile }}"
  …
```

Since I wouldn’t add Gemfile.lock to the repository, `ruby/setup-ruby` will automatically install Bundler v2 for all three gemfiles. And since [Rails 4.2](https://rubygems.org/gems/rails/versions/4.2.11.3) is restricted to versions `>= 1.3.0, < 2.0` of Bundler, my tests cannot run.

Would it be possible to change `ruby/setup-ruby` to consider the contents of `$BUNDLE_GEMFILE + ".lock"` before Gemfile.lock (when `$BUNDLE_GEMFILE` exists)?

Notes:

- `BUNDLE_GEMFILE` is mentioned in https://bundler.io/v2.1/man/bundle-config.1.html
- There is a tool for testing with multiple gemfiles, called “Appraisal”. It seems to work by [setting BUNDLE_GEMFILE](https://github.com/thoughtbot/appraisal/blob/586864393e405a67b1457b563a4d5adc99e50e2d/lib/appraisal/command.rb#L36) before each run.

vecerek:
Fixes #73.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -10,7 +10,7 @@ inputs:
     required: false
     default: 'default'
   bundler:
-    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. The default tries Gemfile.lock and otherwise uses latest.'
+    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. In case of none and Gemfile.lock, the version is determined based on the BUNDLE_GEMFILE environment variable. If the variable is not set, it will default to Gemfile.lock. If a bundler option is not provided, it falls back to using "latest".'
     required: false
     default: 'default'
   bundler-cache:
--- a/action.yml
+++ b/action.yml
@@ -10,7 +10,7 @@ inputs:
     required: false
     default: 'default'
   bundler:
-    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. In case of none and Gemfile.lock, the version is determined based on the BUNDLE_GEMFILE environment variable. If the variable is not set, it will default to Gemfile.lock. If a bundler option is not provided, it falls back to using "latest".'
+    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. For Gemfile.lock, the version is determined based on the BUNDLED WITH section from the file ($BUNDLE_GEMFILE || Gemfile).lock. Defaults to Gemfile.lock if it exists and latest otherwise.'
     required: false
     default: 'default'
   bundler-cache:
--- a/index.js
+++ b/index.js
@@ -153,9 +153,9 @@ function setupPath(newPathEntries) {
   core.exportVariable('PATH', [...newPathEntries, ...cleanPath].join(path.delimiter))
 }
 
-function readBundledWithFromGemfileLock() {
-  if (fs.existsSync('Gemfile.lock')) {
-    const contents = fs.readFileSync('Gemfile.lock', 'utf8')
+function readBundledWithFromGemfileLock(path) {
+  if (fs.existsSync(path)) {
+    const contents = fs.readFileSync(path, 'utf8')
     const lines = contents.split(/\r?\n/)
     const bundledWithLine = lines.findIndex(line => /^BUNDLED WITH$/.test(line.trim()))
     if (bundledWithLine !== -1) {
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -153,17 +153,17 @@ function setupPath(newPathEntries) {
   core.exportVariable('PATH', [...newPathEntries, ...cleanPath].join(path.delimiter))
 }
 
-function readBundledWithFromGemfileLock() {
-  if (fs.existsSync('Gemfile.lock')) {
-    const contents = fs.readFileSync('Gemfile.lock', 'utf8')
+function readBundledWithFromGemfileLock(path) {
```
bquorning, line 156: We might want to rename the function to e.g. `readBundledWithFromGemLockfile`.

index.js
```diff
@@ -175,7 +175,8 @@ async function installBundler(bundlerVersionInput, platform, rubyPrefix, engine,
   var bundlerVersion = bundlerVersionInput
 
   if (bundlerVersion === 'default' || bundlerVersion === 'Gemfile.lock') {
```
bquorning, line 177: Does it make sense still to allow `Gemfile.lock` as an option to the `bundler` configuration? Would it be too hard to make it deprecated?

The hard question is: what should happen when someone uses the `Gemfile.lock` configuration AND has a BUNDLE_GEMFILE env variable?

action.yml
```diff
@@ -10,7 +10,7 @@ inputs:
     required: false
     default: 'default'
   bundler:
-    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. The default tries Gemfile.lock and otherwise uses latest.'
+    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. In case of none and Gemfile.lock, the version is determined based on the BUNDLE_GEMFILE environment variable. If the variable is not set, it will default to Gemfile.lock. If a bundler option is not provided, it falls back to using "latest".'
```
eregon, line 13: `none` actually does not install Bundler at all, and does not run any `bundle` command.

This feels too long and very hard to read due to line length, how about:
```yaml
    description: 'The version of Bundler to install. Either none, 1, 2, latest or Gemfile.lock. For Gemfile.lock, the version is determined based on the BUNDLED WITH section from the file ($BUNDLE_GEMFILE || Gemfile).lock. Defaults to Gemfile.lock if it exists and latest otherwise.
```

I think this would better discussed in the README though, where we can explain more intuitively what happens, and have more space: https://github.com/ruby/setup-ruby#bundler
Then this line does not really need to change.

index.js
eregon: The idea is if we change the default we'd still have a way to express the previous default.
So I think `Gemfile.lock` is good enough (`foo/bar.lock` is still "a lockfile for a "Gemfile"), maybe `lockfile` would be better but it seems less intuitive and not worth it.

index.js
eregon: I think it's OK as-is.

action.yml
vecerek: Thanks, @eregon for the review. I applied your suggestion and added a new paragraph to the README in https://github.com/ruby/setup-ruby/pull/74/commits/c23a60ff7419d5c7436ab69feaa08dfa5a7da235.

eregon:
Released as https://github.com/ruby/setup-ruby/releases/tag/v1.42.0

PR #82, opened 2020-09-02
Add support for `gems.rb`/`gems.locked` during `bundle install`.

ioquatix:
Fixes part of #52 - doesn't work with `gems.rb`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -213,14 +213,26 @@ async function installBundler(bundlerVersionInput, platform, rubyPrefix, engine,
 }
 
 async function bundleInstall(platform, engine, version) {
-  if (!fs.existsSync('Gemfile')) {
-    console.log('No Gemfile, skipping "bundle install" and caching')
-    return
+  if (await bundleInstallSpecific('gems.rb', 'gems.locked', platform, engine, version)) {
+    return true
+  }
+
+  if (await bundleInstallSpecific('Gemfile', 'Gemfile.lock', platform, engine, version)) {
+    return true
+  }
+
+  console.log('No Gemfile/gems.rb, skipping "bundle install" and caching')
+}
+
+async function bundleInstallSpecific(gemsPath, lockPath, platform, engine, version) {
+  if (!fs.existsSync(gemsPath)) {
+    return false
   }
 
   // config
   const path = 'vendor/bundle'
-  const hasGemfileLock = fs.existsSync('Gemfile.lock');
+  const hasGemfileLock = fs.existsSync(lockPath)
+
   if (hasGemfileLock) {
     await exec.exec('bundle', ['config', '--local', 'deployment', 'true'])
   }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -213,14 +213,26 @@ async function installBundler(bundlerVersionInput, platform, rubyPrefix, engine,
...
+  }
+
+  if (await bundleInstallSpecific('Gemfile', 'Gemfile.lock', platform, engine, version)) {
+    return true
+  }
+
+  console.log('No Gemfile/gems.rb, skipping "bundle install" and caching')
+}
+
+async function bundleInstallSpecific(gemsPath, lockPath, platform, engine, version) {
```
eregon, line 227: `gemsPath` is a confusing name, how about `gemfilePath`?

eregon (commented):
It looks like you missed this case:
https://github.com/ruby/setup-ruby/blob/a2ee959e496fb0bbfd6c8f8b64b2f049238cfc56/index.js#L178

eregon:
Thanks for the PR.
Please let me review before merging since I'm the maintainer.

eregon:
Is there any reason to use `gems.rb` over `Gemfile`?
Any adoption of that? It seems extremely few projects use it.

I'm not looking to forward to support every possible experimental Bundler feature, due to the cost to maintain that.
It seems Rails also considers it as a rather pointless change: https://github.com/rails/rails/issues/31295

index.js
```diff
@@ -213,14 +213,26 @@ async function installBundler(bundlerVersionInput, platform, rubyPrefix, engine,
 }
 
 async function bundleInstall(platform, engine, version) {
-  if (!fs.existsSync('Gemfile')) {
-    console.log('No Gemfile, skipping "bundle install" and caching')
-    return
+  if (await bundleInstallSpecific('gems.rb', 'gems.locked', platform, engine, version)) {
```
eregon, line 216: `bundleInstallSpecific` does not `return` in the successful case.
Does the `if` work here? It doesn't seem so:
```
> async function foo() { console.log("foo"); }
> async function call() { var r=await foo(); console.log(r); if (r) { console.log("OK") } else { console.log("NOPE") } }
undefined
> call()
foo
Promise { <pending> }
> undefined
NOPE
```

eregon:
I'll revert this, let's review & discuss it properly before merging.

ioquatix:
I’m already using it so please don’t revert it. Why don’t we just fix the issues you have. I tested both cases in my own repos and it seemed to work fine.

ioquatix:
I am using master.

https://github.com/ioquatix/bake/runs/1064255388?check_suite_focus=true

eregon:
I see, sorry about that.
Using `@master` is expected to be unstable: https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#compatibility

ioquatix:
Sure, I completely understand your POV, but I think it warranted just a tiny bit more discussion, given how easy it is to make the changes you requested.

ioquatix:
https://github.com/ruby/setup-ruby/pull/84

PR #83, opened 2020-09-03
Fix interpolation in addVCVARSEnv() and various cleanups

eregon:
Follow-up of #64
@MSP-Greg Could you review?


eregon:
I'll merge (but would still like your review), because I want to get a release with the fixes from #64 out, and I [checked](https://github.com/ruby/setup-ruby/pull/64#discussion_r482906636) the double `Path` issue is fixed by this PR. The rest is just minor cleanups.

MSP-Greg:
LGTM.  Sorry about that braces/paren thing. Between PowerShell. cmd, Bash, Ruby, JS, I will make similar mistakes again...

PR #92, opened 2020-10-03
Add Ruby 2.7.2

eregon:
Ruby Installer builds are not yet available (https://github.com/oneclick/rubyinstaller2/issues/193).
@larskanis Any ETA on that?
I'm unsure if I should merge this first (but then `2.7` resolves to different versions in different platforms, and there is [some intended incompatibility](https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/) between 2.7.1 and 2.7.2) or wait for the Windows builds (seems better).

cc @chrisseaton


chrisseaton:
> I'm unsure if I should merge this first

I think CI systems like this need to be aggressive about updating, because until CI updates, nobody can update their applications.

eregon:
My concern is if some gem behaves differently on 2.7.1 and 2.7.2 then the CI might fail e.g. only on Windows, which sounds like a platform-specific issue but it actually isn't. And that would cause confusion.
I guess people can use `ruby-version 2.7.1` to workaround that if there is such an issue, but it's not ideal.

Hopefully less warnings in 2.7.2 doesn't break much, but still I wouldn't be surprised if some code depends on the warnings somehow.

chrisseaton:
Can you make 2.7.2 available, but make 2.7 default to 2.7.1 on both platforms for the moment?

MSP-Greg:
Maybe just wait until Lars responds, then decide?

Consider releasing when all platforms are available?  If they aren't, will that affect people pinning to patch/teeny versions, and is that a concern?

eregon:
> Can you make 2.7.2 available, but make 2.7 default to 2.7.1 on both platforms for the moment?

That might cause confusion in why `2.7` doesn't pick the latest 2.7.x release, and make it unclear how the versions are resolved.

Another alternative would be to merge this to master, but not release yet and wait for the Windows build for the release.
Then if someone wants to try 2.7.2 ASAP, one could do `- uses: ruby/setup-ruby@master`.
Of course, that is less stable than `@v1`, but it seems inevitable for such cases as this one.

> Maybe just wait until Lars responds, then decide?

Yeah, I'd like to do that.

> will that affect people pinning to patch/teeny versions?

No, it will not, when using `ruby-version: 2.7.1` it will of course always pick 2.7.1.
Only changes if someone uses `ruby-version: 2.7` or `ruby-version: 2`.

larskanis:
Regarding RubyInstaller-2.7.2 I'm partly blocked by repo.msys2.org [being down currently](https://github.com/msys2/MINGW-packages/issues/7084). Although it's possible to alter the mirrorlist, there are several places to be modified and I don't want to distribute a modified list in the release package. My plan is to make a release until 2020-10-06 either with or without repo.msys2.org.

eregon:
@larskanis Thanks for the reply. Since that seems a prime example of potentially having to wait for something we don't control, I'll merge this to master, but not make a release yet.

Then at least those who want to try 2.6.2 right away on non-Windows can do so (see [here](https://github.com/ruby/setup-ruby/pull/92#issuecomment-703233882)).

I plan to wait for the Windows build to make a new release (and bump `v1`).
Tuesday 6 is just 2 days away, so that sounds great.

chrisseaton:
Thanks this is working for me.

eregon:
Windows build added, and release is done: https://github.com/ruby/setup-ruby/releases/tag/v1.47.0

GRoguelon:
> Windows build added, and release is done: https://github.com/ruby/setup-ruby/releases/tag/v1.47.0
> So now one can simply `- uses: ruby/setup-ruby@v1` to get 2.7.2.

Hi @eregon, I tried with the both `v1` and `master` and my workflow on Github action is failing with: `Version 2.7.2 not found`. Is there a way to see which version of setup-ruby my workflow is using?

eregon:
@GRoguelon it's `- uses: ruby/setup-ruby@v1`, not `- uses: actions/setup-ruby@v1`.

PR #118, opened 2020-12-04
Fixup bundler version for rubies with bundler as a default gem

### LINKED ISSUE #117: `ruby-head` is coming with bundler-2 out of the box, so `bundler: 1` is basically ignored
```
Installing Bundler
  /home/runner/.rubies/ruby-head/bin/gem install bundler -v ~> 1 --no-document
  Successfully installed bundler-1.17.3
  1 gem installed
  Took   0.72 seconds
bundle install
  /home/runner/.rubies/ruby-head/bin/bundle config --local path vendor/bundle
  /home/runner/.rubies/ruby-head/bin/bundle lock
  Fetching gem metadata from https://rubygems.org/...........
  Fetching gem metadata from https://rubygems.org/.
  Resolving dependencies........................................................................
  Bundler could not find compatible versions for gem "bundler":
    In Gemfile.rails-3.x:
      rails (~> 3.0) was resolved to 3.2.15.rc3, which depends on
        bundler (~> 1.0)
  
    Current Bundler version:
      bundler (2.2.0.rc.2)
  This Gemfile requires a different version of Bundler.
  Perhaps you need to update Bundler by running `gem install bundler`?
```

ojab:
Basically use specific bundler version until lockfile is created. Fixes #117, test run which previously failed can be found [here](https://github.com/ojab/watir-rails/runs/1500872128).

### DIFF
--- a/index.js
+++ b/index.js
@@ -230,6 +230,8 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
     const gem = path.join(rubyPrefix, 'bin', 'gem')
     await exec.exec(gem, ['install', 'bundler', '-v', `~> ${bundlerVersion}`, '--no-document'])
   }
+
+  core.exportVariable('BUNDLER_VERSION', bundlerVersion)
 }
 
 async function bundleInstall(gemfile, lockFile, platform, engine, version) {
--- a/dist/index.js
+++ b/dist/index.js
@@ -51402,6 +51402,8 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
     const gem = path.join(rubyPrefix, 'bin', 'gem')
     await exec.exec(gem, ['install', 'bundler', '-v', `~> ${bundlerVersion}`, '--no-document'])
   }
+
+  core.exportVariable('BUNDLER_VERSION', bundlerVersion)
 }
 
 async function bundleInstall(gemfile, lockFile, platform, engine, version) {


eregon:
The failed log is actually https://github.com/ojab/watir-rails/runs/1500383889?check_suite_focus=true
```
  /opt/hostedtoolcache/Ruby/2.6.3/x64/bin/bundle lock
  Fetching gem metadata from https://rubygems.org/...........
  Fetching gem metadata from https://rubygems.org/.
  Resolving dependencies...
  Bundler could not find compatible versions for gem "bundler":
    In Gemfile.rails-4.x:
      rails (~> 4.0) was resolved to 4.2.11.3, which depends on
        bundler (>= 1.3.0, < 2.0)
  
    Current Bundler version:
      bundler (2.1.4)
  This Gemfile requires a different version of Bundler.
  Perhaps you need to update Bundler by running `gem install bundler`?
  
  Could not find gem 'bundler (>= 1.3.0, < 2.0)', which is required by gem 'rails
  (~> 4.0)', in any of the sources.
  
  Bundler could not find compatible versions for gem "rails":
    In Gemfile.rails-4.x:
      rails (~> 4.0)
  
      watir-rails was resolved to 2.2.2, which depends on
        rails (>= 3)
```

package.json
```diff
@@ -26,7 +26,8 @@
     "@actions/core": "^1.2.6",
     "@actions/exec": "^1.0.3",
     "@actions/io": "^1.0.2",
-    "@actions/tool-cache": "^1.3.4"
+    "@actions/tool-cache": "^1.3.4",
+    "js-yaml": "^3.14.0"
```
eregon, line 30: Adding such an heavy dependency for such an edge case is not acceptable.

eregon:
To clarify, this only affects Ruby 2.7+ which ships with Bundler 2, when there is no lockfile, and `bundler: 1` is specified.

index.js
```diff
@@ -235,19 +253,19 @@ async function bundleInstall(gemfile, lockFile, platform, engine, version) {
   // config
   const path = 'vendor/bundle'
 
-  await exec.exec('bundle', ['config', '--local', 'path', path])
+  await exec.exec('bundle', [`_${bundlerVersion}_`, 'config', '--local', 'path', path])
```
eregon, line 256: It's unfortunate that the `_version_` syntax doesn't support `_~> 1_` or so.

OTOH, adding these extra `_version_` in the action output doesn't seem nice, I'd like to keep commands as simple and close to what a user would do as possible. So if we took this approach, I'd want to only add `_version_` when needed (on Ruby 2+ if bundlerVersion == '1').

But, there seems to be a much simpler way:
```
$ BUNDLER_VERSION=1 bundle version
Bundler version 1.17.3 (2018-12-27 commit d7089abb6)
$ BUNDLER_VERSION=2 bundle version
Bundler version 2.2.0.rc.2 (2020-10-07 commit d7c0467653)
```
That relies on `Gem::BundlerVersionFinder`, but that's probably good enough.

Another way would be to hardcode Bundler 1 to be 1.17.3, but it seems less nice than just setting `BUNDLER_VERSION` in the env for the commands that need it.

index.js
```diff
@@ -216,17 +218,33 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
...
+
+  let bundlerSpecification = ''
+
+  const options = {
+    listeners: {
+      stdout: (data) => bundlerSpecification += data.toString()
+    }
+  }
+
+  await exec.exec(gem, ['specification', 'bundler', '-v', `~> ${bundlerVersion}`, 'version'], options)
```
eregon, line 239: This would take extra time for all executions of the action, for a corner case, so that's not ideal.
But I think there is a better solution, see my other comments.

package.json
ojab: Yep, I've done it via `.split`/`.slice` at first, but decided that correctness is better (it adds < 100kb to the current 2.1mb). I'll redo it via `.split/.slice` if we'll decide to merge it.

eregon: The `BUNDLER_VERSION` version doesn't need any parsing, so that seems better.

index.js
ojab: Oh, didn't knew about `BUNDLER_VERSION=`. I'll check it.

ojab:
`BUNDLER_VERSION` works fine \o/ https://github.com/ojab/watir-rails/runs/1503360760?check_suite_focus=true

index.js
```diff
@@ -205,6 +205,8 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
     throw new Error(`Cannot parse bundler input: ${bundlerVersion}`)
   }
 
+  core.exportVariable('BUNDLER_VERSION', bundlerVersion)
```
eregon, line 208: This looks much much simpler :)

I'm not sure we should export `BUNDLER_VERSION` for the whole workflow, but OTOH it's probably clearer that way and it should not hurt.

Can you move this line at the end of this function?
Notably it needs to consider `bundlerVersion` being assigned below.

ojab: ouch, missed that it's assigned below. Fixed.

eregon (approved):
Thanks for the PR!

PR #120, opened 2020-12-07
Pass BUNDLER_VERSION only to commands that need it and not for the whole workflow

ojab:
Looks like we can't delete environment variable and we cannot set envvar for the single step from JS, so let's just set requested bundler version to envvar if it's set, so
```yaml
- use: ruby/setup-ruby@v1
  with:
    ruby-version: 2.7
  env:
    BUNDLER_VERSION: 1
```
will do the trick.

### DIFF
--- a/index.js
+++ b/index.js
@@ -63,12 +63,12 @@ export async function setupRuby(options = {}) {
   if (inputs['bundler'] !== 'none') {
     const [gemfile, lockFile] = detectGemfiles()
 
-    await common.measure('Installing Bundler', async () =>
+    const bundlerVersion = await common.measure('Installing Bundler', async () =>
       installBundler(inputs['bundler'], lockFile, platform, rubyPrefix, engine, version))
 
     if (inputs['bundler-cache'] === 'true') {
       await common.measure('bundle install', async () =>
-          bundleInstall(gemfile, lockFile, platform, engine, version))
+          bundleInstall(gemfile, lockFile, platform, engine, version, bundlerVersion))
     }
   }
 
@@ -231,10 +231,10 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
     await exec.exec(gem, ['install', 'bundler', '-v', `~> ${bundlerVersion}`, '--no-document'])
   }
 
-  core.exportVariable('BUNDLER_VERSION', bundlerVersion)
+  return bundlerVersion
 }
 
-async function bundleInstall(gemfile, lockFile, platform, engine, version) {
+async function bundleInstall(gemfile, lockFile, platform, engine, rubyVersion, bundlerVersion) {
   if (gemfile === null) {
     console.log('Could not determine gemfile path, skipping "bundle install" and caching')
     return false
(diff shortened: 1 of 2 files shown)


eregon:
https://github.com/actions/toolkit/blob/d9347d4ab99fd507c0b9104b2cf79fb44fcc827d/packages/exec/src/interfaces.ts#L5 seems to allow env vars, could you use that instead?
Only the `bundle config` and `bundle lock` commands need it, the other `bundle ...` commands will pick it up from Gemfile.lock.

ojab:
Done, it works: https://github.com/ojab/watir-rails/actions/runs/406505935

MSP-Greg:
I haven't worked much with all the Bundler options and haven't done any serious node work lately.

But, rather than passing the entire ENV to `exec.exec`, won't it pick it up if we use:
```js
proces.env.BUNDLER_VERSION = bundlerVersion
```

Can't recall if `exec.exec` sees changes in the ENV added by the calling code?

ojab:
Yep, works as well, pushed.

eregon:
> But, rather than passing the entire ENV to `exec.exec`, won't it pick it up if we use:
> 
> ```js
> proces.env.BUNDLER_VERSION = bundlerVersion
> ```

That's global state, I think it's much more hacky.

Is there any actual issue with passing a copy of the env? (performance is not relevant for that, the copy is orders of magnitudes faster than the subcommand).

MSP-Greg:
> That's global state, I think it's much more hacky.

Only for the actions process and sub-processes.  Has no effect on subsequent steps?

eregon:
> Only for the actions process and sub-processes. Has no effect on subsequent steps?

It's global state for the current process, and it makes it less clear which commands need the extra env var.
e.g. with async/await you could end up with some other function which unintentionally has the env var set.
Anyway, global state is bad in general, and I'm not willing to compromise on this unless there is an actual issue in terms of usability.

@ojab Could you restore the version passing the env explicitly?

ojab:
Sure, done.

eregon (approved):
Thank you for the fix!

index.js
```diff
@@ -231,31 +231,32 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
...
-async function bundleInstall(gemfile, lockFile, platform, engine, version) {
+async function bundleInstall(gemfile, lockFile, platform, engine, rubyVersion, bundlerVersion) {
   if (gemfile === null) {
     console.log('Could not determine gemfile path, skipping "bundle install" and caching')
     return false
   }
 
   // config
   const path = 'vendor/bundle'
+  const optionsWithBundlerVersion = {options: {env: {...process.env, BUNDLER_VERSION: bundlerVersion}}}
```
eregon, line 245: This should be
```js
const optionsWithBundlerVersion = {env: {...process.env, BUNDLER_VERSION: bundlerVersion}}
```
Fixed in e5a267414fde5bb4dec4599ed01b44941c337575

MSP-Greg:
> with async/await you could end up with some other function which unintentionally has the env var set.

With async, yes, with await, no.  Anyway, 'Apples vs Oranges'.  As long as the unneeded BUNDLER_VERSION is removed...

eregon:
(test) I checked and `bundler: 1` seems to be respected: https://github.com/eregon/setup-ruby/runs/1513198879?check_suite_focus=true

Issue #123, opened 2020-12-11
Locking to a specific bundler version


jdelStrother:
It seems like we can only specify Bundler 1 or 2. Any chance of being able to supply a more specific version? We're having some issues with Bundler 2.2 (https://github.com/rubyjs/libv8/issues/310) so want to lock to 2.1.4 for now

dentarg:
From https://github.com/rubygems/rubygems/issues/4123#issuecomment-743049519 it sounds like rubygems / bundler itself should take care of using the correct Bundler version (needs Ruby version > 2.5 though). What version of Ruby are you using @jdelStrother? Any public build you can share?

deivid-rodriguez:
Yes, that is correct although currently it only works only if multiple versions are installed. Say, if your lock file has `BUNDLED WITH 2.1.4` and both 2.1.4 and 2.2.0 are installed, then 2.1.4 will be chosen. However, if there's only a single version available, we'll default to that version.

I think what's happening in this case is that `ruby/setup-ruby` will install the latest version, not the exact version, so that version will get run since it's the only one installed.

As a workaround, you can specifiy `bundler: none` and then install the proper version yourself by parsing the version from the lock file.

```
gem install bundler -v $(ruby -e 'puts File.read("Gemfile.lock").split("BUNDLED WITH").last.strip')
```

deivid-rodriguez:
Also, for completeness, we had decided to remove the version switching logic from bundler/rubygems and I was planning to get to that after the 2.2.0 release, but this unintended incompatibilities make me hesitate a lot.

My current proposal for rubygems/bundler would be that `bundle install` automatically picks up the `BUNDLED WITH` version, installs that, and then installs the rest of the bundle using the `BUNDLED WITH` version. Then at that point `bundle exec` would always succeed and use the version specified in the `BUNDLED WITH` section of the lockfile.

Regardless of what we end up doing in bundler/rubygems,maybe `ruby/setup-ruby` could provide

```
bundler: none
bundler: exact
bundler: latest
```

so that users can customize this behaviour?

deivid-rodriguez:
Actually there's `bundler: Gemfile.lock` that maybe does this already? :thinking:

jdelStrother:
It looks like it just uses the major version (ie 1 or 2) from the lockfile : https://github.com/ruby/setup-ruby/blob/084885eaca2e353b0f4475be646d25d7d4dd1e89/index.js#L175-L191

eregon:
[...]

PR #126, opened 2020-12-11
Document the `bundler: "x.y.z"` feature

jdelStrother:
It looks like we can specify the bundler version to install with, eg, `bundler: "2.1.4"`, though the documentation implies that we can only install the latest version of bundler 1 or bundler 2.  How about this documentation tweak?

https://github.com/ruby/setup-ruby/issues/123

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -11,7 +11,7 @@ inputs:
     default: 'default'
   bundler:
     description: |
-      The version of Bundler to install. Either 'none', 'latest', 'Gemfile.lock', or a version number (eg 1, 2, "2.1.4").
+      The version of Bundler to install. Either 'none', 'latest', 'Gemfile.lock', or a version number (e.g., 1, 2, 2.1.4).
       For 'Gemfile.lock', the version is determined based on the BUNDLED WITH section from the file Gemfile.lock, $BUNDLE_GEMFILE.lock or gems.locked.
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -11,7 +11,7 @@ inputs:
     default: 'default'
   bundler:
     description: |
-      The version of Bundler to install. Either 'none', 1, 2, 'latest' or 'Gemfile.lock'.
+      The version of Bundler to install. Either 'none', 'latest', 'Gemfile.lock', or a version number (eg 1, 2, "2.1.4").
```
eregon, line 14:
```suggestion
      The version of Bundler to install. Either 'none', 'latest', 'Gemfile.lock', or a version number (e.g., 1, 2, 2.1.4).
```

eregon (commented):
This is fine, but we should test that use case with a new job or workflow in CI.
So far it seems to work, but it's not tested properly so it might not work well.

eregon:
There are actually a lot of checks for `bundlerVersion === '1'` (or `=== '2'`), so we'll need to review those before claiming it's supported.

eregon:
I'll merge this commit manually together with the fixes, thanks for the PR.

eregon:
Integrated as part of https://github.com/ruby/setup-ruby/pull/127

PR #127, opened 2020-12-11
Allow an exact Bundler version, and use the exact same version if present in lockfile

### LINKED ISSUE #123: Locking to a specific bundler version

eregon:
* Fixes #123


eregon:
CI seems fine, I'll merge and release this: https://github.com/eregon/setup-ruby/runs/1537440141?check_suite_focus=true

PR #135, opened 2020-12-28
Update windows-version.js with 3.0.0, add Windows 3.0.0 to CI

MSP-Greg:
Thanks @larskanis & @eregon


windows-versions.js
```diff
@@ -1,36 +1,37 @@
...
-  "2.6.5": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.6.5-1/rubyinstaller-2.6.5-1-x64.7z",
-  "2.6.6": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.6.6-1/rubyinstaller-2.6.6-1-x64.7z",
-  "2.7.0": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.7.0-1/rubyinstaller-2.7.0-1-x64.7z",
-  "2.7.1": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.7.1-1/rubyinstaller-2.7.1-1-x64.7z",
-  "2.7.2": "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-2.7.2-1/rubyinstaller-2.7.2-1-x64.7z",
-  "head": "https://github.com/oneclick/rubyinstaller2/releases/download/rubyinstaller-head/rubyinstaller-head-x64.7z",
-  "mingw": "https://github.com/MSP-Greg/ruby-loco/releases/download/ruby-master/ruby-mingw.7z",
-  "mswin": "https://github.com/MSP-Greg/ruby-loco/releases/download/ruby-master/ruby-mswin.7z"
-}
+export const versions = {
```
eregon, line 1: It looks like the line endings for this file changed, because we see every line as "new".
Probably I should use `File.binwrite` for Windows.
I'll try to push a fix.

eregon (approved):
Thanks!

windows-versions.js
MSP-Greg: Good catch.  I was frustrated with minitest & ruby-head...

I switched to a new desktop last week recently.  I may have something misconfigured...

One reason I got the new desktop is WLS2/Ubuntu.

Already building Ruby master on it.  Takes less than 3 minutes for build & test, test-spec is ~ 7 sec.

PR #139, opened 2021-01-05
install specific bundler version when minor and patch versions are defined

### LINKED ISSUE #138: Newer Bundler patch version is installed
I noticed that not the exact Bundler version specified in Gemfile.lock is installed, but a newer patch version, if available.

In my case, I am using Bundler 2.2.3, then 2.2.4 is released, installed by `setup-ruby`, and then Rspec complains because it expects 2.2.3.

pdf file might be nil · uplinkhq:app@f39eba7 2021-01-04 17-49-03

I think the fix should be to run `gem install bundler -v 2.2.3` instead of `gem install bundler -v ~> 2.2.3`, so that the exact version is installed.

Happy to submit a PR! 😸

manuelmeurer:
fixes #138

### DIFF
--- a/index.js
+++ b/index.js
@@ -229,7 +229,8 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
     console.log(`Using Bundler 1 shipped with ${engine}`)
   } else {
     const gem = path.join(rubyPrefix, 'bin', 'gem')
-    await exec.exec(gem, ['install', 'bundler', '-v', `~> ${bundlerVersion}`, '--no-document'])
+    const bundlerVersionConstraint = bundlerVersion.match(/^\d+\.\d+\.\d+/) ? bundlerVersion : `~> ${bundlerVersion}`
+    await exec.exec(gem, ['install', 'bundler', '-v', bundlerVersionConstraint, '--no-document'])
   }
 
   return bundlerVersion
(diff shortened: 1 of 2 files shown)


manuelmeurer:
Should I do the change in dist/index.js as well or is it somehow autogenerated on release?

manuelmeurer:
@eregon Are those failures relevant to my change?
TruffleRuby on Ubuntu 18.04 reports an `OutOfMemoryError`...

index.js
```diff
@@ -229,7 +229,8 @@ async function installBundler(bundlerVersionInput, lockFile, platform, rubyPrefi
     console.log(`Using Bundler 1 shipped with ${engine}`)
   } else {
     const gem = path.join(rubyPrefix, 'bin', 'gem')
-    await exec.exec(gem, ['install', 'bundler', '-v', `~> ${bundlerVersion}`, '--no-document'])
+    const bundlerVersionConstraint = bundlerVersion.match(/^\d+\.\d+\.\d+$/) ? bundlerVersion : `~> ${bundlerVersion}`
```
eregon, line 232: I think we should remove the `$` here so e.g. `2.2.0.rc.2` would also be considered an exact version.
(https://rubygems.org/gems/bundler/versions)

eregon (commented):
Looks great, just one small refinement.

eregon:
> @eregon Are those failures relevant to my change?
> TruffleRuby on Ubuntu 18.04 reports an `OutOfMemoryError`...

Not caused by your change, it's a known issue, it's being worked on.

index.js
ciityweb: @eregon Done!

manuelmeurer: @eregon Done!

eregon (approved):

Issue #128, opened 2020-12-12
Known third-party issues


eregon:
* Ruby 2.3 + Bundler 2: https://github.com/rubygems/rubygems/issues/3570
  * Bundler 1 is used instead on Ruby 2.3 automatically in ruby/setup-ruby (Ruby 2.3 is EOL and uses a very old RubyGems)
* JRuby 9.1 + Bundler 2: https://github.com/ruby/setup-ruby/issues/108
  * Bundler 1 is used  instead on JRuby 9.1 automatically in ruby/setup-ruby (Ruby 2.3 is EOL and uses a very old RubyGems)

---

Fixed upstream, simply use a newer Bundler release:
* JRuby + Bundler 2.1.4 + Windows + `bundle exec rake`: https://github.com/rubygems/rubygems/issues/3256 and https://github.com/ruby/setup-ruby/issues/18, fixed in Bundler 2.2.0
* JRuby + Bundler 2.2.0 + Windows: https://github.com/rubygems/rubygems/issues/4129, fixed in Bundler 2.2.2
* Bundler 2.2.0 + Gemfile.lock created with older version: https://github.com/rubygems/rubygems/issues/4123, fixed in Bundler 2.2.1

deivid-rodriguez:
For completeness, the fix for rubygems/rubygems#3570 was released with bundler 2.2.0.rc.1, and regarding #108 we never got a report about it :shrug:.

eregon:
Regarding #108, I was hoping @JonRowe would report it to Bundler or JRuby in https://github.com/ruby/setup-ruby/issues/108#issuecomment-726009156, but I probably wasn't clear enough. @JonRowe Could you do that?

Maybe I should write this in the issue template: only file a bug here after you've filed a bug upstream, but it's hard to make a general rule that's still correct in all cases.

deivid-rodriguez:
Yeah, no problem at all! It might be a bug in jruby, but since bundler 1 works, I guess we could probably do something to workaround it. I think it could be due to us stopping using `Dir.chdir` there and instead the `:chdir` parameter of `Kernel.system` and family.

If that was the issue, we could easily workaround it by using `git -C` which I'm sure has been supported for many many years.

eregon:
Marking this issue as closed as there is outstanding 3rd party issue, that is not already worked around.

PR #146, opened 2021-01-26
Remove bundler enforcement on ruby 2.3

bpinto:
The fix for rubygems/rubygems#3570 was released with bundler 2.2.0.rc.1.

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -72,9 +72,6 @@ export async function installBundler(bundlerVersionInput, lockFile, platform, ru
   if (engine === 'ruby' && rubyVersion.match(/^2\.[012]/)) {
     console.log('Bundler 2 requires Ruby 2.3+, using Bundler 1 on Ruby <= 2.2')
     bundlerVersion = '1'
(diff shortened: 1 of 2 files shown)


eregon:
Related: https://github.com/ruby/setup-ruby/issues/128

I guess the reason for the change is you want to use Bundler 2 on Ruby 2.3?
Seems reasonable now that that bug is fixed, but I would not be too surprised if there are other issues with Bundler on Ruby 2.3.

You need to update the generated files as mentioned in CONTRIBUTING.md.

bpinto:
> I guess the reason for the change is you want to use Bundler 2 on Ruby 2.3?

You are correct. 👍 

> You need to update the generated files as mentioned in CONTRIBUTING.md.

Sorry, I read that when I was creating the PR and after running `npm run package` I forgot to push the updated commit 😓

eregon (approved):
Thanks for the PR

eregon:
Released as https://github.com/ruby/setup-ruby/releases/tag/v1.64.0

PR #152, opened 2021-02-02
Support bundle install --local

mlarraz:
If the cache path exists (where *.gem files get stored), use offline mode.

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -148,8 +148,18 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
     console.log(`Found cache for key: ${cachedKey}`)
   }
 
+  const installFlags = [
+    '--jobs', '4',
+  ]
+
+  // Use binary cache if available
+  const vendorCachePath = process.env['BUNDLE_CACHE_PATH'] || 'vendor/cache'
+  if (fs.existsSync(vendorCachePath)) {
+    installFlags.push('--local')
+  }
+
   // Always run 'bundle install' to list the gems
-  await exec.exec('bundle', ['install', '--jobs', '4'])
+  await exec.exec('bundle', ['install', ...installFlags])
 
   // @actions/cache only allows to save for non-existing keys
   if (cachedKey !== key) {


eregon:
`vendor/cache` could exist for another purpose, or not have all the necessary gems, I think this solution is not really reliable.
I think it would need to be an explicit opt-in rather.

What happens if this `--local` flag is not specified, does Bundler use an existing `vendor/cache` if there is?

Also in this setup, I guess you need to check in in `git` or somehow fetch the `vendor/cache`?
It won't be able to initialize itself, right?

What's the advantage of using `bundler-cache: true` in this situation? I guess to avoid recompiling C extensions?

mlarraz:
> What happens if this `--local` flag is not specified, does Bundler use an existing `vendor/cache` if there is?

I believe it will fall back on it, but still need to check with rubygems.org to see if any of the gems have platform-specific extensions that should be downloaded. `--local` skips this step and assumes the correct files are in `vendor/cache`.

> Also in this setup, I guess you need to check in `*.gem` in `git` or somehow fetch the `vendor/cache`?
> It won't be able to initialize itself, right?

Correct. This can be bootstrapped by running `bundle package` (https://bundler.io/v2.0/man/bundle-package.1.html) which downloads all the `*.gem` files into `vendor/cache`.

Afterwards running `bundle install` with any lockfile changes will add/remove the correct `*.gem` files in that directory. The expectation is that `vendor/cache` be committed anytime `Gemfile.lock` is.

> What's the advantage of using `bundler-cache: true` in this situation? I guess to avoid recompiling C extensions?

The existing `bundler-cache: true` option caches all the files and extensions for the gems in a bundle (in `vendor/bundle`), avoiding compilation.

However, it still requires connecting to rubygems.org (or any other sources) when `bundle install` is run, and downloading missing gems. With sufficiently large Gemfiles, this can become a big bottleneck. `vendor/cache` supplements `vendor/bundle` by eliminating the need for a network connection when `Gemfile.lock` has changed. Projects can and do use both at the same time.

bundler.js
```diff
@@ -148,8 +148,18 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
     console.log(`Found cache for key: ${cachedKey}`)
   }
 
+  const installFlags = [
+    '--jobs', '4',
+  ]
+
+  // Use binary cache if available
+  const vendorCachePath = process.env['BUNDLE_CACHE_PATH'] || 'vendor/cache'
```
mlarraz, line 156: This mirror's Bundler own configuration (as documented [here](https://bundler.io/v2.2/man/bundle-config.1.html#LIST-OF-AVAILABLE-KEYS))

eregon:
> However, it still requires connecting to rubygems.org (or any other sources) when `bundle install` is run, and downloading missing gems. With sufficiently large Gemfiles, this can become a big bottleneck. `vendor/cache` supplements `vendor/bundle` by eliminating the need for a network connection when `Gemfile.lock` has changed.

Can you share some numbers here to quantify that? I would expect resolving gems and downloading gems is not significantly slower than downloading a larger repository or the `vendor/cache` of `.gem` files from somewhere else. If resolving is slow, it sounds more like a Bundler issue.

eregon:
Closing as this PR seems inactive. We can reopen based on a reply to my comment above.

PR #157, opened 2021-03-05
Remove quotations for 3.0

okuramasafumi:
Version numbers are not string elsewhere, but only 3.0 is a string.
This is a bit confusing so quotations has been removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


eregon:
It needs to be, see https://github.com/actions/runner/issues/849.
It would be great if you can leave a comment there.
IMHO it should be fixed in GitHub Actions, but until then we should use a String for `3.0` to workaround.

okuramasafumi:
Oh I see. Then should it be commented on README so that no one sends similar PR?

eregon:
Yes, that's a good idea, could you make a PR for that? :)

PR #158, opened 2021-03-05
Add a comment about version quotations to README

okuramasafumi:
As described in the comment, due to a bug discussed on https://github.com/actions/runner/issues/849,
we currently have to use quotes for  '3.0' as a workaround.
This change makes it easy for people not to remove quotes as they misunderstand it's a typo.


eregon (approved):
Thanks!

PR #164, opened 2021-03-24
JRuby 9.3 (head) and forward ships bundler

enebo:
JRuby 9.3 will ship with bundler and jruby-head already does.  This PR, like truffleruby, will not bother installing bundler since it is already installed.

I was confused about one thing.  When JRuby 9.3 is released how do we change this action to say bundler is already installed?  ruby-version appears to only report 'jruby' and 'jruby-head' to this JS function.  I would like to make sure when we do put out that version we are not again reinstalling bundler.

### DIFF
--- a/common.js
+++ b/common.js
@@ -60,6 +60,8 @@ export function isBundler2Default(engine, rubyVersion) {
     return isHeadVersion(rubyVersion) || floatVersion(rubyVersion) >= 2.7
   } else if (engine === 'truffleruby') {
     return isHeadVersion(rubyVersion)
+  } else if (engine === 'jruby') {
+    return isHeadVersion(rubyVersion)
   } else {
     return false
   }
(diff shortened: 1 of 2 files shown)


eregon (approved):
Looks good, thanks.
When there is a 9.3 release we should update that logic to use something like the `=== 'ruby'` logic for 9.3.

eregon:
I added 9.3 to return true there too: https://github.com/ruby/setup-ruby/commit/5ee10197cf484c79397f113605e24f9b5f3ff885

However this function alone doesn't decide whether to install Bundler or not.
It's a bit more complicated than that, I forgot: dddacd7ec16cae1a8190380a744ef47cb2aa0bed
So for releases, we still install the latest Bundler release, which I think helps to get the latest Bundler fixes.

enebo:
@eregon Thanks amazingly I had that in my original PR but I could not figure out how rubyVersion would get the 9.3.  tbh part of me does not want to know :)   Thanks for the additional tweak!

PR #175, opened 2021-05-14
cache version option

### LINKED ISSUE #174: Feature request: don't cache specified gems and/or clear whole bundler cache
Using setup-ruby with `bundler-cache: true` I now have in the cache a gem with native extensions, which are "broken" (successfully compiled, but against a set of C libraries other than I need). It seems to be impossible to replace contents of the cache. Even if Gemfile.lock changes, gems already present in the cache are reused and only missing ones installed, which is usually extremely useful, but not when you desperately need to rebuild the broken gem.

Would it be possible to add a new option, accepting a list of gems and excluding them from caching? E. g.

```yaml
# ...
- uses: ruby/setup-ruby@v1
  with:
    ruby-version: 3.0.0
    bundler-cache: true
    #bundler-cache-exclude: # the new option I dream about
    #  - mysql2
    #  - nokogiri

    # EDIT: turns out that only simple string values are supported as action inputs
    bundler-cache-exclude: mysql2 nokogiri
```

For gems which are not self-contained, but depend on the build environment (like all gems with native extensions) this would occasionally be very useful.

If not, would it be possible to provide at least some mechanism for clearing the whole cache?

igneus:
attempt to fix #174

implements the solution proposed in https://github.com/ruby/setup-ruby/issues/174#issuecomment-839625379

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,20 @@ When there is no lockfile, one is generated with `bundle lock`, which is the sam
 In other words, it works exactly like `bundle install`.
 The hash of the generated lockfile is then used for caching, which is the only correct approach.
 
+#### Dealing with a corrupted cache
+
+In some rare scenarios (like using gems with C extensions whose functionality depends on libraries found on the system
+at the time of the gem's build) it may be necessary to ignore contents of the cache and get and build all the gems anew.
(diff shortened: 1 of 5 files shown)


eregon (approved):
This looks great, thank you for the PR

PR #188, opened 2021-06-09
Add JRuby 9.2.18.0

walro:

PR #190, opened 2021-06-10
Enable SassC test on JRuby on Windows

walro:
9.2.18.0 has been released, so this should work now.

Follow-up to #188


eregon (approved):
Thanks!

PR #195, opened 2021-06-17
Update package json, fix error from @actions/io update

MSP-Greg:
2nd commit - `io.mkdirP` now raises an error if the directory exists


MSP-Greg:
I just noticed that ncc is now 0.28.6.  Can that be updated?  If so, I'll revise...

eregon:
Yeah, that sounds fine to update as long as it keeps working.

ruby-builder.js
```diff
@@ -44,7 +45,9 @@ async function downloadAndExtract(platform, engine, version, rubyPrefix) {
   const parentDir = path.dirname(rubyPrefix)
 
   await io.rmRF(rubyPrefix)
-  await io.mkdirP(parentDir)
+  if (!fs.existsSync(parentDir)) {
```
eregon, line 48:
```suggestion
  if (!fs.statSync(parentDir).isDirectory()) {
```

Could you change to that?
So we only skip if it exists and it's a directory, otherwise we let mkdirP handle the error.

MSP-Greg: I don't think so...

First, from `const parentDir = path.dirname(rubyPrefix)`, we know that `parentDir` cannot be a file name.  I don't think the method makes any assumptions as to existence, but it does strip the last path segment, so it shouldn't be a file.

After that statement, `parentDir` must be a valid directory path, but it may not exist.  Given that, `fs.existsSync(parentDir)` works as expected.

Conversely, if `parentDir` does not exist, `fs.statSync(parentDir)` fails, because one can't stat a non-existent directory or file.

The error caused by the updates was Windows JRuby only, calling `io.mkdirP("D:\\")`, which is a valid directory on Windows Images.

Note that as a 'root' directory, it cannot be created, even if it doesn't exist, e.g., the statement `io.mkdirP("G:\\")` should fail...

eregon: > First, from `const parentDir = path.dirname(rubyPrefix)`, we know that `parentDir` cannot be a file name. I don't think the method makes any assumptions as to existence, but it does strip the last path segment, so it shouldn't be a file.

`rubyPrefix` is however a path we compute, not necessirely a path on the filesystem.
So dirname(/foo/bar/baz) = /foo/bar but doesn't say anything about whether /foo/bar is on the filesystem or its file type.

> The error caused by the updates was Windows JRuby only, calling `io.mkdirP("D:\\")`, which is a valid directory on Windows Images.

It sounds like this is an unintended regression, could you report it upstream?

> Conversely, if `parentDir` does not exist, `fs.statSync(parentDir)` fails, because one can't stat a non-existent directory or file.

I see, we could do this then:

```js
if (!(fs.existsSync(parentDir) && fs.statSync(parentDir).isDirectory())) {
```

MSP-Greg: I'll make the above change, but I don't see how it could be a file.

> It sounds like this is an unintended regression, could you report it upstream?

If you ask a bunch of coders what should happen if you try to create a directory that already exists, I think quite a few would say that an error should be raised.  Hence, I'd rather leave it as is.

As to whether it's a bug fix or could be considered a breaking change, that's complicated...

Since you merge the new JRuby version, I need to rebase...

eregon: > If you ask a bunch of coders what should happen if you try to create a directory that already exists, I think quite a few would say that an error should be raised. Hence, I'd rather leave it as is.

Not for `mkdir -p`. `mkdir -p` should be no-op if it already exists.

> As to whether it's a bug fix or could be considered a breaking change, that's complicated...

There is no question, it is an incompatibility since it works before and it breaks after.
Please report it. Then they can decide whether to fix it.

> Since you merge the new JRuby version, I need to rebase...

Yes, sorry about that, I thought I already merged that PR.

MSP-Greg: Thanks.  I guess I left some 'read between the lines'.  Also, I started coding on the first pc's (IBM & Compaq), so I'm not just familiar with Windows, I'm familiar with DOS.

The problem with `mkdir` and Windows is that a drive is not equivalent to a directory.  I'll create an issue mentioning that `io.mkdirP("D:\\")` fails on Windows...

eregon (approved):
Thanks for the update!

MSP-Greg:
Rebased.  I'll redo the ucrt PR after the merge.

Obviously, you can squash, wasn't sure if you'd want the separate commits....

PR #196, opened 2021-06-19
Runtime - Add logging of both PATH removals & additions, CI - improve 'PATH' step

MSP-Greg:
Commits:

**'Improve Path info in CI, add workflow_dispatch'** - only changes Actions, shows 'PATH' step info with each PATH segment on an individual line, display Windows Path in a normal window style.

**'Update runtime logging of Path adjustments, show additions'** - the runtime logging had a group named 'Cleaning Path'.  It only showed path deletions.  Changed name to 'Modifying Path', and added logging of path additions.

### DIFF
--- a/common.js
+++ b/common.js
@@ -154,17 +154,17 @@ export function setupPath(newPathEntries) {
   const originalPath = process.env[envPath].split(path.delimiter)
   let cleanPath = originalPath.filter(entry => !/\bruby\b/i.test(entry))
 
+  core.startGroup(`Modifying ${envPath}`)
+
   // First remove the conflicting path entries
   if (cleanPath.length !== originalPath.length) {
-    core.startGroup(`Cleaning ${envPath}`)
-    console.log(`Entries removed from ${envPath} to avoid conflicts with Ruby:`)
+    console.log(`Entries removed from ${envPath} to avoid conflicts with default Ruby:`)
     for (const entry of originalPath) {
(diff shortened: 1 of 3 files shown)


MSP-Greg:
Some thing is wrong with Actions.  My fork always stopped on 'Test installing a Gemfile with nokogiri on TruffleRuby'

common.js
```diff
@@ -176,5 +176,13 @@ export function setupPath(newPathEntries) {
   } else {
     newPath = newPathEntries
   }
+  console.log(`Entries added to ${envPath} to use selected Ruby:`)
+  for (const entry of newPath) {
+    if (!cleanPath.includes(entry)) {
```
eregon, line 181: I think we don't need this check, right?
Because those entries are prepended to the PATH, even if they were already in it.

eregon (approved):
Looks good, thanks!

common.js
eregon: I'll remove that check, easy enough

MSP-Greg: Sorry, AFK (sleeping), common at 6:40 A on a weekend.

> `if (!cleanPath.includes(entry)) {`

I don't normally just add 'stuff'.  Can't remember why I had the conditional.  As you noticed, it's kind of pointless under expected/normal conditions, and should be removed, as it may hide image changes.  We should see everything added to Path, since the Actions logging doesn't show `addPath` items...  Thanks

PR #197, opened 2021-07-02
Windows: Use "ridk enable" like variable setup and add UCRT

larskanis:
This PR adds a "ridk enable" like setup of environment variables. "ridk enable" and "ridk exec" are the preferred ways to enable MSYS2 in RubyInstaller2. This not only sets the PATH but also several other variables that are required for pkgconf, autotools, etc.

"ridk enable" mimics the variable setup of MSYS2 "bash --login", but enables powershell and cmd.exe to run MSYS2 commands.

As a fallback, in case the RubyInstaller mechanism fails, the PATH is extended to use MSYS64 with MINGW or UCRT, but no other variables are set.

The PR installs gcc when running on Windows-UCRT, since there is no UCRT capable compiler preinstalled on github actions. That might be a temporary solution until @MSP-Greg finds another way to package gcc in a faster install solution.

A Github Action run on Windows looks like so: https://github.com/larskanis/ruby-pg/runs/2972095461?check_suite_focus=true#step:3:19

### DIFF
--- a/common.js
+++ b/common.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 const util = require('util')
 const stream = require('stream')
 const crypto = require('crypto')
+const child_process = require('child_process')
 const core = require('@actions/core')
 const { performance } = require('perf_hooks')
 
@@ -150,7 +151,7 @@ export function win2nix(path) {
 }
 
 // JRuby is installed after setupPath is called, so folder doesn't exist
-function rubyIsUCRT(path) {
+export function rubyIsUCRT(path) {
   return !!(fs.existsSync(path) &&
     fs.readdirSync(path, { withFileTypes: true }).find(dirent =>
       dirent.isFile() && dirent.name.match(/^x64-ucrt-ruby\d{3}\.dll$/)))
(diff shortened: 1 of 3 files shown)


MSP-Greg:
> until @MSP-Greg finds another way to package gcc in a faster install solution

I finished most of the work last night, and it is faster.  Another repo builds a 7z file of the ucrt install, and using it considerably shortens the install time.  I'll post a message in #193.

common.js
```diff
@@ -168,13 +176,27 @@ export function setupPath(newPathEntries) {
...
+      // Same as "ridk enable" on the command line
+      const envbuf = child_process.execFileSync(`${newPathEntries[0]}\\ruby`, ['-rruby_installer/runtime', '-e', 'puts RubyInstaller::Runtime.msys2_installation.enable_msys_apps_per_cmd'])
+      const envvars = envbuf.toString().trim().split(/\r?\n/)
+
+      envvars.forEach( (envvar) => {
+        console.log(`SET ${envvar}`)
+        const parts = envvar.split("=", 2)
+        core.exportVariable(parts[0], parts[1])
+      })
+    } catch (ex) {
```
eregon, line 192: I think we should at minimum log the exception here to help debug if anything goes wrong.
I think more reliable would be to have no fallback, do we need this fallback?

eregon:
I think this has been superseded https://github.com/ruby/setup-ruby/pull/224 which was merged, so I'll close.

PR #198, opened 2021-07-02
Windows: Install rubyinstaller-head version with Devkit

larskanis:
This is another attempt to install UCRT tools. It downloads `rubyinstaller-devkit-head-x64.exe` and installs ruby with all bundled MSYS2 and UCRT64 tools. The rest is equal to #197.

This is a very simple way to get the UCRT64 environment running, but since it installs not just UCRT64 tools but also MSYS64, it takes between 40 to 50 seconds for Ruby+MSYS2.

See here in action: https://github.com/larskanis/setup-ruby/runs/2976045933?check_suite_focus=true#step:3:13

### DIFF
--- a/common.js
+++ b/common.js
@@ -4,6 +4,7 @@ const fs = require('fs')
 const util = require('util')
 const stream = require('stream')
 const crypto = require('crypto')
+const child_process = require('child_process')
 const core = require('@actions/core')
 const { performance } = require('perf_hooks')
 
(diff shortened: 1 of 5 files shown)


eregon:
I like this approach because it's fairly simple, and probably closer to what typical Ruby users on Windows would do (i.e., I guess most download rubyinstaller with DevKit).
If the GH Windows images don't have enough, then installing the DevKit too makes sense to me.

> since it installs not just UCRT64 tools but also MSYS64

Is that required, or something part of the archive/exe but we don't need it / redundant with what's on the GH Windows images?

larskanis:
@eregon MSYS2 and MINGW/UCRT aren't separate components so far - that's why more files are installed than necessary. But it's simple to make them distinct components and install only the missing part. I'll try it out.

larskanis:
I tried to add an option to rubyinstaller to install only MINGW/UCRT tools and re-use MSYS2 from another path, but MSYS2 is fixated about the path of MINGW tools. Working with links between MSYS2 and MINGW makes things too complicated for being used only in the CI environment. On the other hand installing the whole RubyInstaller with Devkit takes more than twice the time of extracing just the necessary files from 7z. So I think we should proceed with the mechanism described by @MSP-Greg in https://github.com/ruby/setup-ruby/pull/194#issuecomment-906552246 .

PR #199, opened 2021-07-07
Add Ruby 2.7.4

rbclark:
https://www.ruby-lang.org/en/news/2021/07/07/ruby-2-7-4-released/


JonRowe:
`2.6.8` and `3.0.2` have also been released, all three are CVE fix releases

eregon:
I'll use https://github.com/ruby/setup-ruby/pull/200 as that has the three versions.

PR #200, opened 2021-07-07
Add Ruby 2.6.8, 2.7.4 and 3.0.2

eregon:


JonRowe:
Thanks for getting these up so quickly @eregon

PR #202, opened 2021-07-15
Suggest absolute path for BUNDLE_GEMFILE

jonathanhefner:
This ensures the gemfile can be found when a step changes the working directory.

---

A little while back, I encountered an issue when setting up CI for Rails gems and using the `bin/test` script that Rails generates.  This addition fixed it.  I had written myself a reminder to submit this PR, but, unfortunately, I didn't write down any further details.  But I hope this helps someone.


eregon (approved):
Thanks!

Issue #204, opened 2021-07-26
Ruby 1.9 support


mullermp:
Hello! I am wondering if it's possible to support Ruby 1.9. It doesn't appear to be supported (available versions are 2.0+). Some of our projects unfortunately still support Ruby 1.9 and we'd like to test against it where possible.

The last release is here: https://www.ruby-lang.org/en/news/2014/11/13/ruby-1-9-3-p551-is-released/

eregon:
Sorry, no, 1.9 is too old, I think it's high time to drop support for it.
You can try to build it with https://github.com/ruby/ruby-builder if you want to try.

You can potentially use Docker images if you need such ancient Ruby versions.

mullermp:
Unfortunately we can't drop support for it - we own the AWS SDK For Ruby and customers (even internal) still use it. Just to push back a bit - why support Ruby 2.0 or 2.1 in `setup-ruby` then - aren't those old too?

mullermp:
I can create a PR if that helps. I think it's worthwhile to support as long as you are supporting Ruby 2.0, 2.1, 2.2, 2.3 etc like other end of life rubies.

eregon:
Ruby 2 is a symbolic barrier that I think is worth making it clear it's the strict minimum version.
Consider 2.0 - 2.3 as only partially supported, e.g, a system package (e.g. database) linking against the system libssl (only libssl 1.1 available on modern OS) would not work with them since they need libssl 1.0 which is out of life, in other words they are insecure and they have known security issues.

1.8.7 is likely impossible to build on any modern OS.
1.9 not sure, you are welcome to try and report the results.

mullermp:
I agree on having a symbolic barrier for a strict minimum version - you have to draw the line somewhere! My opinion is that 1.9.3 should be that barrier because IIRC it broke some syntax from 1.8.7 but 2.0.0 is the same as 1.9.3 with added new backward compatible capabilities (such as keyword arguments). I also think that it uses the same OpenSSL as 2.0 to 2.3.

mullermp:
I got 1.9.3 to build! I had to fork a number of packages. On [this pull request](https://github.com/mullermp/ruby-builder/pull/1/checks) you can see that build ruby step succeeds however the CI fails on installing json.

Here's what I did:

[...]

PR #205, opened 2021-07-26
Support Ruby 1.9

mullermp:
See https://github.com/ruby/setup-ruby/issues/204


mullermp:
I wasn't sure about dist/index.js. I made the change manually but it seems to be a generated file.

eregon:
There is quite a bit more work to get 1.9 to run in this action's CI, so I'll make my own PR

PR #210, opened 2021-07-31
Keep Dependencies Up-to-date 

mmorel-35:
This PR Allows You to receive pr for each update on ruby, npm and Github-actions dependencies. Your workflow are passed so when it doesn't fails you stay up-to-date on a click.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,7 @@ updates:
   directory: /
   schedule:
     interval: weekly
+- package-ecosystem: npm
+  directory: /
+  schedule:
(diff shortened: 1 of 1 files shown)


eregon:
What's the advantage of this?
Updating Bundler can't work, there is no Gemfile.lock, and the Gemfile is used for tests, they are not dependencies.
Updating GitHub actions seems pointless, they already point to the latest stable version.

So that leaves npm, which would then be updated weekly?
I'm not interested to have a PR every week just because a dependency has a new release, or to make a release just for that every week.

PR #224, opened 2021-10-08
Windows - Add support for ucrt & Windows-2022 image

MSP-Greg:
See commits.  On Windows-2022 image, code adds the MSYS2 packages from 'base-devel' group needed for Ruby builds.

Also adds ENV variables by running `ridk enable`, testing for the file's existence first.  Uses code similar to the code for mswin's vcvars script.  Parses the output of the set command, so it's not dependent on ridk's output.

Windows 2019 Image - No changes when using with a mingw Ruby. When using a ucrt Ruby, a pre-built ucrt 7z file is installed from https://github.com/MSP-Greg/setup-msys2-gcc/releases/tag/msys2-gcc-pkgs.  The 7z files are generated three times a day in that repo.

Windows 2022 Image - Since the image has no MSYS2 build tools installed, both an MSYS2 package and a gcc package (mingw or ucrt) are installed from the MSP-Greg/setup-msys2-gcc repo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,8 @@ jobs:
         - { os: windows-2016, ruby: mingw }
         - { os: windows-2019, ruby: mingw }
         - { os: windows-2019, ruby: mswin }
+        - { os: windows-2022, ruby: mingw }
+        - { os: windows-2022, ruby: head }
         exclude:
         - { os: windows-2016, ruby: 1.9 }
         - { os: windows-2016, ruby: debug }
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,15 +15,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11.0, windows-2016, windows-2019 ]
+        os: [ ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11.0, windows-2016, windows-2019, windows-2022 ]
         ruby: [ 1.9, '2.0', 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, ruby-head, jruby, jruby-head, truffleruby, truffleruby-head, truffleruby+graalvm, truffleruby+graalvm-head ]
         include:
         - { os: windows-2016, ruby: mingw }
         - { os: windows-2019, ruby: mingw }
         - { os: windows-2019, ruby: mswin }
         - { os: windows-2022, ruby: mingw }
-        - { os: windows-2022, ruby: ucrt }
-        - { os: windows-2022, ruby: head }
+        - { os: windows-2022, ruby: ucrt  }
         exclude:
         - { os: windows-2016, ruby: 1.9 }
         - { os: windows-2016, ruby: debug }
@@ -37,6 +36,12 @@ jobs:
         - { os: windows-2019, ruby: truffleruby-head }
         - { os: windows-2019, ruby: truffleruby+graalvm }
         - { os: windows-2019, ruby: truffleruby+graalvm-head }
+        - { os: windows-2022, ruby: 1.9 }
+        - { os: windows-2022, ruby: debug }
+        - { os: windows-2022, ruby: truffleruby }
+        - { os: windows-2022, ruby: truffleruby-head }
+        - { os: windows-2022, ruby: truffleruby+graalvm }
+        - { os: windows-2022, ruby: truffleruby+graalvm-head }
 
     name: ${{ matrix.os }} ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,7 @@ jobs:
         - { os: windows-2019, ruby: mingw }
         - { os: windows-2019, ruby: mswin }
         - { os: windows-2022, ruby: mingw }
+        - { os: windows-2022, ruby: ucrt }
         - { os: windows-2022, ruby: head }
         exclude:
         - { os: windows-2016, ruby: 1.9 }
--- a/index.js
+++ b/index.js
@@ -51,6 +51,13 @@ export async function setupRuby(options = {}) {
   createGemRC(engine, version)
   envPreInstall()
 
+  // JRuby can use compiled extension code, so make sure gcc exists.
+  // As of Jan-2022, JRuby compiles against msvcrt.
+  if (platform.startsWith('windows') && (engine === 'jruby') && 
+    !fs.existsSync('C:\\msys64\\mingw64\\bin\\gcc.exe')) {
+    await require('./windows').installJRubyTools()
+  }
+
   const rubyPrefix = await installer.install(platform, engine, version)
 
   // When setup-ruby is used by other actions, this allows code in them to run
--- a/windows.js
+++ b/windows.js
@@ -50,13 +50,51 @@ export async function install(platform, engine, version) {
 
   let toolchainPaths = (version === 'mswin') ? await setupMSWin() : await setupMingw(version)
 
-  common.setupPath([`${rubyPrefix}\\bin`, ...toolchainPaths])
-
   if (!inToolCache) {
     await downloadAndExtract(engine, version, url, base, rubyPrefix);
   }
 
+  const winMSYS2Type = common.setupPath([`${rubyPrefix}\\bin`, ...toolchainPaths])
+
+  const virtualEnv = common.getVirtualEnvironmentName()
+
+  if (( winMSYS2Type === 'ucrt64') || !['windows-2019', 'windows-2016'].includes(virtualEnv)) {
+    await installGCCTools(winMSYS2Type)
+
+    if (!['windows-2019', 'windows-2016'].includes(virtualEnv)) {
+      await installMSY2Tools()
+    }
+  }
+
   return rubyPrefix
+}
+
+// Actions windows-2022 image does not contain any mingw or ucrt build tools.  Install tools for it,
+// and also install ucrt tools on earlier versions, which have msys2 and mingw tools preinstalled.
+async function installGCCTools(type) {
+  const downloadPath = await common.measure(`Download ${type} build tools`, async () => {
+    let url = `https://github.com/MSP-Greg/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/${type}.7z`
+    console.log(url)
+    return await tc.downloadTool(url)
+  })
+
+  await common.measure(`Extracting ${type} build tools`, async () =>
+    // -aoa overwrite existing, -bd disable progress indicator
+    exec.exec('7z', ['x', downloadPath, '-aoa', '-bd', '-oC:\\msys64'], { silent: true }))
+}
+
+// Actions windows-2022 image does not contain any MSYS2 build tools.  Install tools for it.
+// A subset of the MSYS2 base-devel group
+async function installMSY2Tools() {
+  const downloadPath = await common.measure(`Download msys2 build tools`, async () => {
+    let url = `https://github.com/MSP-Greg/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/msys2.7z`
+    console.log(url)
+    return await tc.downloadTool(url)
+  })
+
+  await common.measure(`Extracting msys2 build tools`, async () =>
+    // -aoa overwrite existing, -bd disable progress indicator
+    exec.exec('7z', ['x', downloadPath, '-aoa', '-bd', '-oC:\\msys64'], { silent: true }))
 }
 
 async function downloadAndExtract(engine, version, url, base, rubyPrefix) {
@@ -68,7 +106,7 @@ async function downloadAndExtract(engine, version, url, base, rubyPrefix) {
   })
 
   await common.measure('Extracting Ruby', async () =>
-    exec.exec('7z', ['x', downloadPath, `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
+    exec.exec('7z', ['x', downloadPath, '-bd', `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
 
   if (base !== path.basename(rubyPrefix)) {
     await io.mv(path.join(parentDir, base), rubyPrefix)
--- a/windows.js
+++ b/windows.js
@@ -22,6 +22,8 @@ const certFile = 'C:\\Program Files\\Git\\mingw64\\ssl\\cert.pem'
 const msys = `${drive}:\\DevKit64`
 const msysPathEntries = [`${msys}\\mingw\\x86_64-w64-mingw32\\bin`, `${msys}\\mingw\\bin`, `${msys}\\bin`]
 
+const virtualEnv = common.getVirtualEnvironmentName()
+
 export function getAvailableVersions(platform, engine) {
   if (engine === 'ruby') {
     return Object.keys(rubyInstallerVersions)
@@ -57,8 +59,6 @@ export async function install(platform, engine, version) {
   }
 
   const winMSYS2Type = common.setupPath([`${rubyPrefix}\\bin`, ...toolchainPaths])
-
-  const virtualEnv = common.getVirtualEnvironmentName()
 
   if (!['windows-2019', 'windows-2016'].includes(virtualEnv)) {
     await installMSY2Tools()
@@ -175,9 +175,22 @@ async function setupMSWin() {
 /* Sets MSVC environment for use in Actions
  *   allows steps to run without running vcvars*.bat, also for PowerShell
  *   adds a convenience VCVARS environment variable
- *   this assumes a single Visual Studio version being available in the windows-latest image */
+ *   this assumes a single Visual Studio version being available in the window images */
 export function addVCVARSEnv() {
-  const vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
+  let vcVars = ''
+  switch (virtualEnv) {
+    case 'windows-2016':
+      vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
+      break
+    case 'windows-2019':
+      vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
+      break
+    case 'windows-2022':
+      vcVars = '"C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
+      break
+    default:
+      throw new Error(`Unknown Windows Image: ${virtualEnv}`)
+  }
   core.exportVariable('VCVARS', vcVars)
 
   let newEnv = new Map()
--- a/windows.js
+++ b/windows.js
@@ -134,6 +134,11 @@ async function downloadAndExtract(engine, version, url, base, rubyPrefix) {
 async function setupMingw(version) {
   core.exportVariable('MAKE', 'make.exe')
 
+  // rename these to avoid confusion when Ruby is using OpenSSL 1.0.2
+  // most current extconf files look for 1.1.x dll files first, which is the
+  // version of the renamed files
+  if (common.floatVersion(version) <= 2.4) { renameSystem32Dlls() }
+
   if (common.floatVersion(version) <= 2.3) {
     core.exportVariable('SSL_CERT_FILE', certFile)
     await common.measure('Installing MSYS', async () => installMSYS(version))
@@ -217,6 +222,17 @@ export function addVCVARSEnv() {
     }
   }
   return newPathEntries
+}
+
+// ssl files cause issues with non RI2 Rubies (<2.4) and ruby/ruby's CI from
+// build folder due to dll resolution
+function renameSystem32Dlls() {
+  const sys32 = 'C:\\Windows\\System32\\'
+  const badFiles = ['libcrypto-1_1-x64.dll', 'libssl-1_1-x64.dll']
+  badFiles.forEach( (bad) => {
+    let fn = `${sys32}${bad}`
+    if (fs.existsSync(fn)) { fs.renameSync(fn, `${fn}_`) }
+  })
 }
 
 // Sets MSYS2 ENV variables set from running `ridk enable`
--- a/windows.js
+++ b/windows.js
@@ -77,13 +77,13 @@ export async function install(platform, engine, version) {
 // Actions windows-2022 image does not contain any mingw or ucrt build tools.  Install tools for it,
 // and also install ucrt tools on earlier versions, which have msys2 and mingw tools preinstalled.
 async function installGCCTools(type) {
-  const downloadPath = await common.measure(`Download ${type} build tools`, async () => {
+  const downloadPath = await common.measure(`Downloading ${type} build tools`, async () => {
     let url = `https://github.com/MSP-Greg/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/${type}.7z`
     console.log(url)
     return await tc.downloadTool(url)
   })
 
-  await common.measure(`Extracting ${type} build tools`, async () =>
+  await common.measure(`Extracting  ${type} build tools`, async () =>
     // -aoa overwrite existing, -bd disable progress indicator
     exec.exec('7z', ['x', downloadPath, '-aoa', '-bd', '-oC:\\msys64'], { silent: true }))
 }
@@ -91,13 +91,13 @@ async function installGCCTools(type) {
 // Actions windows-2022 image does not contain any MSYS2 build tools.  Install tools for it.
 // A subset of the MSYS2 base-devel group
 async function installMSY2Tools() {
-  const downloadPath = await common.measure(`Download msys2 build tools`, async () => {
+  const downloadPath = await common.measure(`Downloading msys2 build tools`, async () => {
     let url = `https://github.com/MSP-Greg/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/msys2.7z`
     console.log(url)
     return await tc.downloadTool(url)
   })
 
-  await common.measure(`Extracting msys2 build tools`, async () =>
+  await common.measure(`Extracting  msys2 build tools`, async () =>
     // -aoa overwrite existing, -bd disable progress indicator
     exec.exec('7z', ['x', downloadPath, '-aoa', '-bd', '-oC:\\msys64'], { silent: true }))
 }
@@ -110,7 +110,7 @@ async function downloadAndExtract(engine, version, url, base, rubyPrefix) {
     return await tc.downloadTool(url)
   })
 
-  await common.measure('Extracting Ruby', async () =>
+  await common.measure('Extracting  Ruby', async () =>
     exec.exec('7z', ['x', downloadPath, '-bd', `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
 
   if (base !== path.basename(rubyPrefix)) {
--- a/windows.js
+++ b/windows.js
@@ -35,6 +35,10 @@ export function getAvailableVersions(platform, engine) {
 export async function install(platform, engine, version) {
   const url = rubyInstallerVersions[version]
 
+  // The windows-2016 and windows-2019 images have MSYS2 build tools (C:/msys64/usr)
+  // and MinGW build tools installed.  The windows-2022 image has neither.
+  const hasMSYS2PreInstalled = ['windows-2019', 'windows-2016'].includes(virtualEnv)
+
   if (!url.endsWith('.7z')) {
     throw new Error(`URL should end in .7z: ${url}`)
   }
@@ -62,11 +66,13 @@ export async function install(platform, engine, version) {
 
   // install msys2 tools for all Ruby versions, only install mingw or ucrt for Rubies >= 2.4
 
-  if (!['windows-2019', 'windows-2016'].includes(virtualEnv)) {
+  if (!hasMSYS2PreInstalled) {
     await installMSYS2Tools()
   }
 
-  if ((( winMSYS2Type === 'ucrt64') || !['windows-2019', 'windows-2016'].includes(virtualEnv)) &&
+  // windows 2016 and 2019 need ucrt64 installed, 2022 and future images need
+  // ucrt64 or mingw64 installed, depending on Ruby version
+  if (((winMSYS2Type === 'ucrt64') || !hasMSYS2PreInstalled) &&
     (common.floatVersion(version) >= 2.4)) {
     await installGCCTools(winMSYS2Type)
   }
@@ -112,7 +118,7 @@ async function installMSYS2Tools() {
 }
 
 // Windows JRuby can install gems that require compile tools, only needed for
-// windows-2022 image
+// windows-2022 and later images
 export async function installJRubyTools() {
   await installMSYS2Tools()
   await installGCCTools('mingw64')
@@ -127,6 +133,7 @@ async function downloadAndExtract(engine, version, url, base, rubyPrefix) {
   })
 
   await common.measure('Extracting  Ruby', async () =>
+    // -bd disable progress indicator, -xr extract but exclude share\doc files
     exec.exec('7z', ['x', downloadPath, '-bd', `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
 
   if (base !== path.basename(rubyPrefix)) {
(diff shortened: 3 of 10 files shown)


MSP-Greg:
I created a fork of ruby-loco, and built both mingw & ucrt on WIndows 2022 using this PR branch for setup-ruby.  Both compiled and passed all tests.  This PR runs jobs for mingw & ucrt builds on both Windows 2019 & 2022.

MSP-Greg:
@eregon

Good day.

All the below is Windows specific...

Sorry for all the CI, a few typos, etc.  Since windows-2022 does not have MSYS2 build tools installed (there's pro's and con's for that decision), we need to install the tools here, as any CI that installs an extension gem will fail otherwise.   With the previous images, we just needed to install the tools for the old Rubies (2.3 and earlier), now that's required for all Rubies.

This is working in ruby/ruby, and I've tested it with nokogiri and puma.  nokogiri needs the xml related packages installed, so it's using a branch in setup-ruby-pkgs that pairs with this PR's branch.

I'll check if it works with ruby/openssl, which has an mswin build, and if that's ok, it's ready to go...

bundler.js
```diff
@@ -91,7 +91,11 @@ export async function installBundler(bundlerVersionInput, lockFile, platform, ru
   } else {
     const gem = path.join(rubyPrefix, 'bin', 'gem')
     const bundlerVersionConstraint = /^\d+\.\d+\.\d+/.test(bundlerVersion) ? bundlerVersion : `~> ${bundlerVersion}`
-    await exec.exec(gem, ['install', 'bundler', '-v', bundlerVersionConstraint])
+    if ((bundlerVersionConstraint === '~> 2') && (common.floatVersion(rubyVersion) >= 2.3)) {
+      await exec.exec(gem, ['update', '--system', '--no-document'])
```
eregon, line 95: I guess this is to avoid the error like https://github.com/ruby/setup-ruby/pull/253#issuecomment-1004061786 ?

I dislike updating RubyGems though, even more so for a new release, it can cause more things to break confusingly.

windows.js
```diff
@@ -153,3 +219,36 @@ export function addVCVARSEnv() {
   }
   return newPathEntries
 }
+
+// ssl files cause issues with non RI2 Rubies (<2.4) and ruby/ruby's CI from
+// build folder due to dll resolution
+function renameSystem32Dlls() {
```
eregon, line 225: This seems quite risky, is there any other way to solve the issue?
Is it only for Ruby <= 2.3? It seems done for all cases here.
It seems to work fine before for <= 2.3, no?

.github/workflows/test.yml
```diff
@@ -21,6 +21,14 @@ jobs:
...
         - { os: windows-2019, ruby: mingw }
         - { os: windows-2019, ruby: mswin }
+        - { os: windows-2022, ruby: '2.0' }
+        - { os: windows-2022, ruby: 2.3   }
+        - { os: windows-2022, ruby: 2.4   }
+        - { os: windows-2022, ruby: '3.0' }
+        - { os: windows-2022, ruby: 3.1   }
+        - { os: windows-2022, ruby: mingw }
+        - { os: windows-2022, ruby: ucrt  }
+        - { os: windows-2022, ruby: head  }
```
eregon, line 31: I think we should add `windows-2022` under `os:` so it's tested on all Ruby versions like 2016/2019.
Potentially we can drop windows-2016 at some point so we keep only 2 versions per OS.

windows.js
```diff
@@ -50,15 +54,63 @@ export async function install(platform, engine, version) {
...
 
-  common.setupPath([`${rubyPrefix}\\bin`, ...toolchainPaths])
-
   if (!inToolCache) {
     await downloadAndExtract(engine, version, url, base, rubyPrefix);
   }
 
+  const winMSYS2Type = common.setupPath([`${rubyPrefix}\\bin`, ...toolchainPaths])
+
+  // install msys2 tools for all versions, only install mingw or ucrt for Rubies >= 2.4
```
eregon, line 63:
```suggestion
  // install msys2 tools for all Ruby versions, only install mingw or ucrt for Rubies >= 2.4
```

windows.js
```diff
@@ -126,9 +179,22 @@ async function setupMSWin() {
 /* Sets MSVC environment for use in Actions
  *   allows steps to run without running vcvars*.bat, also for PowerShell
  *   adds a convenience VCVARS environment variable
- *   this assumes a single Visual Studio version being available in the windows-latest image */
+ *   this assumes a single Visual Studio version being available in the window images */
```
eregon, line 182:
```suggestion
 *   this assumes a single Visual Studio version being available in the windows images */
```

bundler.js
MSP-Greg: See https://github.com/rubygems/rubygems/issues/5245.

The most recent Bundler version is a bit more tightly coupled to RubyGems than in the past.  Outside of Windows, there are warnings thrown about using an older version of RubyGems that doesn't support x, y, and z.  I'll see if I can find some examples.

Generally, I don't like doing anything that isn't necessary.

This is the one part of the PR that I added so it would work, but I hope there's a better solution.

**Off-topic:** I need some more coffee.  It's earlier here.

windows.js
MSP-Greg: This is quite involved as to why it's needed.

To be clear, these **are not MSFT files**.  One of the oldest and most popular Windows OpenSSL implementations (built with Visual C) installs the OpenSSL dll's in `C:\Windows\System32`.  Normally, that folder is considered 'off-limits', and should be used only by MSFT.

I've had this in setup-ruby-pkgs for a long time, and it's never caused an issue.  It causes all sorts of WTF issues if it's not done...

windows.js
MSP-Greg: Updated

windows.js
MSP-Greg: Updated with 'Windows'

.github/workflows/test.yml
MSP-Greg: Updated.  I left out mswin on windows-2022.  See https://bugs.ruby-lang.org/issues/18362

MSP-Greg:
Re the windows-2022 JRuby failures:
```
  C:\Windows\system32\tar.exe -xz -C /d/ -f /d/a/_temp/a1828e52-1472-430c-9f5e-f1202ac6f2b7
  tar.exe: Error opening archive: Failed to open '/d/a/_temp/a1828e52-1472-430c-9f5e-f1202ac6f2b7'
```
Previous images had the MSYS2 tar installed, windows-2022 does not, so the MSFT tar.exe is used, but it needs a Windows path.  Looking at it now...

MSP-Greg:
@eregon re the Windows JRuby extraction issue, I've got that fixed.  I just noticed the 'Windows JRuby' step in CI (`gem install sassc`).  It fails on windows-2022 because there are no build tools.

So, should they be installed?  If so, I need to see whether they should be mingw or ucrt?

eregon:
> So, should they be installed? If so, I need to see whether they should be mingw or ucrt?

I guess it doesn't really matter, so the easier the better.
For sassc, a C++ compiler is needed to create a `.dll` and that's then loaded via FFI.

MSP-Greg:
Locally, I built the sassc gem with jruby-head, ruby-mingw, & ruby-ucrt.

Using objdump to see what dll's they load, the following is the data.  JRuby may be compiled with MSFT Visual Studio.  Not sure about enabling that, but if one were to choose between MinGW & UCRT, the MinGW build's dll's line up with JRuby much better.  I'll work on it.
```
———————————————————————————————————————————————— ruby-mingw
ADVAPI32.dll
bcrypt.dll
imagehlp.dll
IPHLPAPI.DLL
KERNEL32.dll
libgmp-10.dll
msvcrt.dll
SHELL32.dll
USER32.dll
DLL Name: WS2_32.dll

———————————————————————————————————————————————— ruby-ucrt
ADVAPI32.dll
api-ms-win-crt-convert-l1-1-0.dll
api-ms-win-crt-environment-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
api-ms-win-crt-locale-l1-1-0.dll
api-ms-win-crt-math-l1-1-0.dll
api-ms-win-crt-private-l1-1-0.dll
api-ms-win-crt-runtime-l1-1-0.dll
api-ms-win-crt-stdio-l1-1-0.dll
api-ms-win-crt-string-l1-1-0.dll
api-ms-win-crt-time-l1-1-0.dll
api-ms-win-crt-utility-l1-1-0.dll
bcrypt.dll
imagehlp.dll
IPHLPAPI.DLL
KERNEL32.dll
libgmp-10.dll
SHELL32.dll
USER32.dll
WS2_32.dll


———————————————————————————————————————————————— MinGW libsass.so
ADVAPI32.dll
KERNEL32.dll
libgcc_s_seh-1.dll
libstdc++-6.dll
msvcrt.dll

———————————————————————————————————————————————— UCRT libsass.so
ADVAPI32.dll
api-ms-win-crt-convert-l1-1-0.dll
api-ms-win-crt-environment-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
api-ms-win-crt-locale-l1-1-0.dll
api-ms-win-crt-math-l1-1-0.dll
api-ms-win-crt-private-l1-1-0.dll
api-ms-win-crt-runtime-l1-1-0.dll
api-ms-win-crt-stdio-l1-1-0.dll
api-ms-win-crt-string-l1-1-0.dll
api-ms-win-crt-time-l1-1-0.dll
KERNEL32.dll
libgcc_s_seh-1.dll
libstdc++-6.dll


———————————————————————————————————————————————— Windows JRuby
ADVAPI32.DLL
KERNEL32.dll
msvcrt.dll
USER32.dll
WS2_32.dll
```

MSP-Greg:
Fixed the issues with JRuby & windows-2022.  Appears there are problems with Actions macOS servers.  Anyway, I think this is complete, with the issue of RubyGems possibly remaining.

BTW, Happy New Year!

MSP-Greg:
@eregon & @larskanis

This morning I decided to look at the Bundler/RubyGems issue, as this PR currently updates RubyGems instead of updating Bundler for `((bundlerVersionConstraint === '~> 2') && (common.floatVersion(rubyVersion) >= 2.3))`.

So, I ran it my fork of this repo, [results](https://github.com/MSP-Greg/ruby-setup-ruby/actions/runs/1654436785).  The only failing jobs are Windows 3.1.0.

There are a few repos running from the PR branch of my fork, Puma being one.  It has been passing using it.

I reverted the RubyGems installation, [results](https://github.com/MSP-Greg/puma/runs/4704367970).  As above, the Windows 3.1.0 jobs failed, but also Ruby 2.5 jobs on **every OS failed**.  The 2.5 failures maybe due to Psych, don't have time to look this morning.

Another 'issue' in both CI runs is that all jobs using Ruby 2.3 thru 2.5 have a warning shown, similar to:
```
Your RubyGems version (2.6.14.4)) has a bug that prevents `required_ruby_version` from working for Bundler
```

So, not sure what to do about this.  Lars, if you have time, maybe see if the issue with the 3.1 bin can be fixed?  I'll try to look later...  Regardless, we've got warnings and failures in all OS's for Ruby 2.3 thru 2.5 when upgrading to Bundler 2.3.4...

MSP-Greg:
I updated RubyGems/Bundler handling.  For Ruby 2.3 thru 2.5 (all EOL), it does `gem update --system`.  For Windows 3.1.0, it installs Bundler with `--force`.  So, it works, no warnings...

MSP-Greg:
@larskanis

Re the bundler issue, it's messy, so I don't think anything can be 'fixed'.  Part of the problem is supplying files so that ruby will run from both a Windows shell and the MSYS2 shell...

MSP-Greg:
@larskanis

Sorry, finally really looked at the issue.  I think all the bash bind files (not *.bat or *.cmd), need the following removed, as it causes a UI prompt when updating bundler.  Same as deleting everything before the shebang line:
```
:""||{ ""=> %q<-*- ruby -*-
@"%~dp0ruby" -x "%~f0" %*
@exit /b %ERRORLEVEL%
};{ #
bindir="${0%/*}" #
exec "$bindir/ruby" "-x" "$0" "$@" #
>,
}
```

.github/workflows/test.yml
eregon: I still see those `include` in the diff, but they seem redundant now

bundler.js
```diff
@@ -91,7 +91,16 @@ export async function installBundler(bundlerVersionInput, lockFile, platform, ru
...
+    const floatVersion = common.floatVersion(rubyVersion)
+    if ((bundlerVersionConstraint === '~> 2') && (floatVersion >= 2.3) && (floatVersion <= 2.5)) {
+      // stop warnings similar to
+      // Your RubyGems version (2.6.14.4)) has a bug that prevents `required_ruby_version` from working for Bundler
+      await exec.exec(gem, ['update', '--system', '--no-document'])
+    } else if (platform.startsWith('windows-') && (floatVersion === 3.1)) {
+      await exec.exec(gem, ['install', 'bundler', '-v', bundlerVersionConstraint, '--force'])
+    } else {
+      await exec.exec(gem, ['install', 'bundler', '-v', bundlerVersionConstraint])
+    }
```
eregon, line 103: Could you drop the changes in this file and rebase to master?

bundler.js
```diff
@@ -91,7 +91,16 @@ export async function installBundler(bundlerVersionInput, lockFile, platform, ru
   } else {
     const gem = path.join(rubyPrefix, 'bin', 'gem')
     const bundlerVersionConstraint = /^\d+\.\d+\.\d+/.test(bundlerVersion) ? bundlerVersion : `~> ${bundlerVersion}`
-    await exec.exec(gem, ['install', 'bundler', '-v', bundlerVersionConstraint])
+    const floatVersion = common.floatVersion(rubyVersion)
+    if ((bundlerVersionConstraint === '~> 2') && (floatVersion >= 2.3) && (floatVersion <= 2.5)) {
+      // stop warnings similar to
+      // Your RubyGems version (2.6.14.4)) has a bug that prevents `required_ruby_version` from working for Bundler
+      await exec.exec(gem, ['update', '--system', '--no-document'])
```
eregon, line 98: We might need to do something about that one, not sure yet, but I think installing `Bundler < 2.3.0` is less risky: https://github.com/ruby/setup-ruby/issues/228#issuecomment-1009156811
Anyway, I think best to not change that in this PR

index.js
```diff
@@ -51,6 +51,11 @@ export async function setupRuby(options = {}) {
   createGemRC(engine, version)
   envPreInstall()
 
+  // 1st conditional may need to change...
+  if (platform.startsWith('windows-202') && engine === 'jruby') {
```
eregon, line 55: Could you do a better detection there, for instance one that checks if some path exists?

windows.js
```diff
@@ -50,15 +54,70 @@ export async function install(platform, engine, version) {
...
-
   if (!inToolCache) {
     await downloadAndExtract(engine, version, url, base, rubyPrefix);
   }
 
+  const winMSYS2Type = common.setupPath([`${rubyPrefix}\\bin`, ...toolchainPaths])
+
+  // install msys2 tools for all Ruby versions, only install mingw or ucrt for Rubies >= 2.4
+
+  if (!['windows-2019', 'windows-2016'].includes(virtualEnv)) {
```
eregon, line 65: A path check here would be better as well, and we can have a comment this is the case on windows-2022

windows.js
```diff
@@ -50,15 +54,70 @@ export async function install(platform, engine, version) {
...
+    await installMSYS2Tools()
+  }
+
+  if ((( winMSYS2Type === 'ucrt64') || !['windows-2019', 'windows-2016'].includes(virtualEnv)) &&
```
eregon, line 69: Extract `!['windows-2019', 'windows-2016'].includes(virtualEnv)` in a variable, for instance `const hasMSYSPreInstalled = ....` (usage here would be `!hasMSYSPreInstalled`)

windows.js
```diff
@@ -50,15 +54,70 @@ export async function install(platform, engine, version) {
...
+    (common.floatVersion(version) >= 2.4)) {
```
eregon, line 70: Is that a condition for MSYS2 vs MSYS1? If so please extract in a variable which name indicates that (e.g., usesMSYS2)

windows.js
```diff
@@ -50,15 +54,70 @@ export async function install(platform, engine, version) {
...
+  }
+
   return rubyPrefix
 }
 
+// Actions windows-2022 image does not contain any mingw or ucrt build tools.  Install tools for it,
+// and also install ucrt tools on earlier versions, which have msys2 and mingw tools preinstalled.
+async function installGCCTools(type) {
+  const downloadPath = await common.measure(`Downloading ${type} build tools`, async () => {
+    let url = `https://github.com/MSP-Greg/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/${type}.7z`
```
eregon, line 86: @hsbt Are we OK to depend on a repository outside the ruby GitHub organization for setting up Ruby on Windows? Or should we move that repo to the ruby organization?
(Of course I trust @MSP-Greg and he is the best person to maintain it)

windows.js
```diff
@@ -67,8 +126,8 @@ async function downloadAndExtract(engine, version, url, base, rubyPrefix) {
     return await tc.downloadTool(url)
   })
 
-  await common.measure('Extracting Ruby', async () =>
-    exec.exec('7z', ['x', downloadPath, `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
+  await common.measure('Extracting  Ruby', async () =>
+    exec.exec('7z', ['x', downloadPath, '-bd', `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
```
eregon, line 130: Could you add above `// -bd disable progress indicator` ?

windows.js
```diff
@@ -126,9 +186,22 @@ async function setupMSWin() {
...
  *   allows steps to run without running vcvars*.bat, also for PowerShell
  *   adds a convenience VCVARS environment variable
- *   this assumes a single Visual Studio version being available in the windows-latest image */
+ *   this assumes a single Visual Studio version being available in the Windows images */
 export function addVCVARSEnv() {
-  const vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
+  let vcVars = ''
+  switch (virtualEnv) {
+    case 'windows-2016':
+      vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
```
eregon, line 194: So before it looks like we were using MSVC 2019 on windows-2016, and now we'll use MSCV 2017 there. Any risk of incompatibility?

windows.js
eregon: My concern here is someone might waste a lot of time if for some reason some part of their workflow would use those files and using setup-ruby has the very unexpected side effect to delete or rename files under `C:\Windows\System32` (which are not provided by Microsoft but still).
I remember some people were already confused due to the more extensive Path changes on Windows by this action, but this goes another level. If we keep it we should have a clear logged message mentioning what it and why it does it.

I think it's better to not do this. If those files are problematic then filing an issue at https://github.com/actions/virtual-environments seems the best way forward.

@larskanis What do you think? Having a 3rd opinion might be helpful here.

eregon (commented):
This looks great, I did a full review now.

windows.js
MSP-Greg: > Any risk of incompatibility?

Actually, it's always been broken, VS 2019 isn't installed on windows-2016.

We don't test it here (mswin & windows-2016).  On Actions, ruby/ruby only builds mswin with windows-2019.  And, the only publicly available built, ruby-loco's mswin, is built with 2019.

Lastly, I think GitHub is removing the windows-2016 image in March?  So, leave it an remove when the image is removed?

eregon: The change is fine, thanks for the details.

MSP-Greg:
I'm working on this.  In particular, the MSYS2 conditionals, as windows-2022 does have a minimal MSYS2 install.

One messy thing when talking about MSYS2 is making clear what's being referred to.  MSYS2 has the following build toolsets, all are subfolders of `C:/msys64`:
```
clang32
clang64
clangarm64
mingw32
mingw64
ucrt64
```

But, there are also tools under `C:/msys64/usr`.  This is where the bash shell files, pacman, and common build tools (autoconf, automake, bison, etc) are located.  Also, there's a compiler chain to build them.  Note that most of these files have a dependency on `msys-2.0.dll`, which probably has a bunch of helper functions for windows.  Code built with the above list of toolsets do not have that dependency.

With windows-2019, the common build tools were installed there.  With windows-2022, all common build tools were removed.

In the past. I may have not been clear as to whether I was referring to MSYS2 (the whole thing) or the collection of files in `C:/msys64/usr`.  I will make that clear from now on.

bundler.js
MSP-Greg: Pre-compiled gems are probably more common with Windows, and the `required_ruby_version` bug has definitely caused problems.  So, I'd rather see the above than `Bundler < 2.3.0`.  There has been a lot of recent work in Bundler/RubyGems, so it would be good to install the most recent version available...

Regardless, I'll remove the `gem update`.

index.js
MSP-Greg: How about
```js
  // JRuby can use compiled extension code, so make sure gcc exists.
  // As of Jan-2022, JRuby compiles against msvcrt.
  if (platform.startsWith('windows') && (engine === 'jruby') && 
    !fs.existsSync('C:\\msys64\\mingw64\\bin\\gcc.exe') {
    await require('./windows').installJRubyTools()
  }
```

bundler.js
eregon: I wonder if the required_ruby_version only affects RubyGems or also bundler, if only rubygems it seems a much smaller issue, anyway best discussed on a separate issue.

index.js
eregon: Yes that looks good, might just want to move that gcc path to `common` if it's used in multiple places

windows.js
MSP-Greg: Below is info from a repo I've got that spits out a bunch of data on Actions Images, taken from the windows-2022 image job.  All of the openssl.exe files have the dll's in their own directory.  The `C:\Windows\System32` files are added by the same installer that creates the files at `C:\Program Files\OpenSSL`.
```
——————————————————————————————————————————————————————————————————————— libcrypto-1_1-x64.dll
C:\mysql\bin\libcrypto-1_1-x64.dll
C:\Windows\System32\libcrypto-1_1-x64.dll
C:\Program Files\OpenSSL\bin\libcrypto-1_1-x64.dll
C:\Program Files\Git\mingw64\bin\libcrypto-1_1-x64.dll
c:\tools\php\libcrypto-1_1-x64.dll

——————————————————————————————————————————————————————————————————————— libssl-1_1-x64.dll
C:\mysql\bin\libssl-1_1-x64.dll
C:\Windows\System32\libssl-1_1-x64.dll
C:\Program Files\OpenSSL\bin\libssl-1_1-x64.dll
C:\Program Files\Git\mingw64\bin\libssl-1_1-x64.dll
c:\tools\php\libssl-1_1-x64.dll

——————————————————————————————————————————————————————————————————————— openssl info
OpenSSL 1.1.1  11 Sep 2018    C:\Program Files\OpenSSL\bin\openssl.exe
OpenSSL 1.1.1l  24 Aug 2021   C:\Program Files\Git\mingw64\bin\openssl.exe
OpenSSL 1.1.1l  24 Aug 2021   C:\Program Files\Git\usr\bin\openssl.exe
```

> My concern here is someone might waste a lot of time

I've worked with quite a few extension gems re Windows, Ruby master/head, and Actions.  Never come across a problem.  It is a confusing problem if Rubies that use `ENV['Path']` for their dll resolution and load them because they're in `C:\Windows\System32`...

.github/workflows/test.yml
MSP-Greg: fixed

bundler.js
MSP-Greg: done

bundler.js
MSP-Greg: I can't recall, but it's a PITA.  I think Bundler, but it uses code in RubyGems?  As I recall, it happens when a pre-compiled gem exists, but doesn't support the Ruby version, which will always happen with Ruby master/head.  But, it installs it anyway...

index.js
MSP-Greg: I haven't added to common.  I think we should really try to minimize windows paths in index.js and common.js?  This being the one except, as I was unaware that people used extension code with FFI and JRuby.  BTW, does TruffleRuby build on Windows?

windows.js
MSP-Greg: One is being used to check whether to update/install `C:/msys64/usr` files, and one is used for `C:/msys64/mingw64` or `C:/msys64/ucrt64` files.  Not sure, but added a constant...

windows.js
MSP-Greg: see above

windows.js
MSP-Greg: One is being used to check whether to update/install `C:/msys64/usr` files, and one is used for `C:/msys64/mingw64` or `C:/msys64/ucrt64` files.  Not sure, but added a constant...

windows.js
MSP-Greg: see above

windows.js
MSP-Greg: done

windows.js
eregon: Could you do this only for RI1, so Ruby <= 2.3, and log a message in that case?
I think it's more reasonable to use workarounds like this for old versions, but for recent versions we should avoid any such hacks.

windows.js
eregon: Is there an easy way to check if some path/dir/file exist to asses if MSYS2 is preinstalled? I would prefer to avoid depending on a given image having MSYS2 installed or not if possible.

index.js
eregon: OK.
>  BTW, does TruffleRuby build on Windows?

No it doesn't.

windows.js
MSP-Greg: I did think about this.  The base MSYS2 installation (bash shell & pacman) is installed on windows-2022.  What is not installed are the 'shared build tools' that are not dependent on the build toolset.  It's very similar to the bash shell that git installs, with the addition of pacman.

From the discussions in actions/virtual-environments, I've gotten the impression that they will lock what MSYS2 packages are installed on images.  IOW, if there are changes, they will not add packages to a pre-existing image, they will only do so when a new image is released.  Also, I doubt they will make changes to the MSYS2 install on new images in the future.

eregon: OK. I think it could be an issue for self-hosted windows, but I guess that's very rare and not supported currently, so fine as-is.

windows.js
eregon: @MSP-Greg could you answer/fix? Also extra space in `( winMSYS2Type` above

windows.js
eregon: I'm fine to merge as-is for now and we can figure that part later.

windows.js
MSP-Greg: Sorry, fixed the space and added a comment...

windows.js
MSP-Greg: Windows OpenSSL dll's have different names, depending on the version, the break happens between 1.0.2 and 1.1.0.  Updated for Ruby 2.4 and earlier, as some extconf files that work with OpenSSL look for the 1.1.0 names first.  Windows Ruby 2.4 uses 1.0.2.  See comment

eregon (approved):
Thank you for this great work!

eregon:
Released: https://github.com/ruby/setup-ruby/releases/tag/v1.92.0

MSP-Greg:
I suspect there are things about TruffleRuby that even without a few nights sleep you would be aware of.  Given how long I've used Ruby with Windows, same thing.  Hence, sorry about not commenting several things.

One change that's contained in [MSP-Greg/setup-msys2-gcc](https://github.com/MSP-Greg/setup-msys2-gcc) is that I included all the packages needed to build Ruby, which is also true for all the extension stdlib/gems in Ruby repos.  I was tired of people using [MSP-Greg/setup-ruby-pkgs](https://github.com/MSP-Greg/setup-ruby-pkgs) to install packages that are required by Ruby.  Also, I thought matching Ubuntu was a good idea.

But, when removing [MSP-Greg/setup-ruby-pkgs](https://github.com/MSP-Greg/setup-ruby-pkgs) from CI here and there, there was a fair bit of code that needed to move to this repo.  And that's where I got a little light with the comments...  Thanks for reviewing this and keeping the repo up to date.  Too many people think it's automatic.

lopopolo:
Hi, is it likely that the MSYS changes mentioned here are interfering with libraries available in the base installed packages on the GHA runner?

It looks like these changes combined with using Ruby 3.1.0 on `windows-2019` add a non-default libclang to the path which breaks Rust bindgen.

Example build: https://github.com/artichoke/artichoke/runs/4811419721?check_suite_focus=true

MSP-Greg:
I haven't looked at the build, but Ruby 3.1 is built with the MSYS2 UCRT64 toolset, and Ruby 3.0 (and previous) were built with the MSYS2 MINGW64 toolset.

Not quite sure what you mean by 'non-default libclang'? Nothing has changed in the handling of Path.

lopopolo:
bindgen does a dynamic load of libclang. this is the runtime error:

```
  thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at C:\\msys64\\mingw64\\bin\\libclang.dll could not be opened: LoadLibraryExW failed"', C:\Users\Default\.cargo\registry\src\github.com-1ecc6299db9ec823\bindgen-0.59.2\src/lib.rs:2144:31
```

MSP-Greg:
When using Ruby 3.1 (or head or ucrt), you'll need to (at a minimum) add the following:
```
pacman -S mingw-w64-ucrt-x86_64-clang
```

The windows-latest (windows-2019) image only contains the mingw64 toolset (maybe also mingw32, but we don't use).  As mentioned above, Windows Ruby 3.1 uses the ucrt64 toolset, which is not pre-installed on any Actions Windows images.

Note that in the windows-2022 image the mingw64 toolset is not installed, nor are the MSYS2 build tools.

MSP-Greg:
@lopopolo

See https://github.com/artichoke/artichoke/pull/1636

MSP-Greg:
@eregon 
Time for a new release?  ruby/ruby mingw CI is failing on a few openssl tests due to the dll resolution issue.  I'm double checking right now.

eregon:
@MSP-Greg There is already https://github.com/ruby/setup-ruby/releases/tag/v1.92.0 and the two commits after are just cleanups, shouldn't matter. What's the issue?

MSP-Greg:
ruby/ruby mingw CI
```
  Retrying...
  
    1) Failure:
  OpenSSL::TestConfig#test_s_parse_include [D:/a/ruby/ruby/src/test/openssl/test_config.rb:158]:
  <["default", "sec-a", "sec-b", "sec-main"]> expected but was
  <["default", "sec-main"]>.
  
    2) Error:
  OpenSSL::TestCipher#test_aes_ccm:
  OpenSSL::Cipher::CipherError: 
      D:/a/ruby/ruby/src/test/openssl/test_cipher.rb:374:in `ccm_data_len='
      D:/a/ruby/ruby/src/test/openssl/test_cipher.rb:374:in `block (2 levels) in new_decryptor'
      D:/a/ruby/ruby/src/test/openssl/test_cipher.rb:374:in `each'
      D:/a/ruby/ruby/src/test/openssl/test_cipher.rb:374:in `block in new_decryptor'
      <internal:kernel>:90:in `tap'
      D:/a/ruby/ruby/src/test/openssl/test_cipher.rb:372:in `new_decryptor'
      D:/a/ruby/ruby/src/test/openssl/test_cipher.rb:187:in `test_aes_ccm'
```

I'm testing right now, I believe the SYSTEM32 dll's need to renamed for all Rubies.

MSP-Greg:
See https://github.com/ruby/setup-ruby/pull/258

windows.js
hsbt: Sorry, late response.

It's okay to merge this. I think it's better to transfer ruby org because we should support them by team members.

MSP-Greg: That's fine with me.  No idea how to do so.  I'll see if I can find anything...

MSP-Greg: Forgot one thing.  There is a folder in the repo, https://github.com/MSP-Greg/setup-msys2-gcc/tree/main/.github/actions/upload-binary.

It contains the js packaged code that uploads the asembled files to the release.  The code is very similar to the code I use for ruby-loco, so it's handled the code for Actions `mingw`, `ucrt`, and `mswin` Windows Rubies.

Let me see where that code is, as updates to the Actions packages normally warrant an update to it...

hsbt: @MSP-Greg Can you transfer them to me ( @hsbt )? After that, I will transfer to ruby organization. And I also create `actions` team and promote you and eregon to admin grant.

MSP-Greg: @hsbt

I apologize for the delay.  I made many changes to both repos, and I wanted to update some docs re newer Rubies, ucrt, mswin, etc.  Some changes to mswin builds are not yet complete, but it's been too long already.

So, I'm ready to transfer them.  Is now a good time?

Thanks, Greg

hsbt: No problem. Can you transfer them to me? After that, I also transfer them to ruby org and I promote you and @eregon to their admin.

hsbt: 👍 

I'm also done. Just available https://github.com/ruby/setup-ruby-pkgs and https://github.com/ruby/setup-msys2-gcc

PR #233, opened 2021-11-24
Update ruby-builder-versions.js with new Ruby releases

rofreg:
[According to CONTRIBUTING.md](https://github.com/ruby/setup-ruby/blob/master/CONTRIBUTING.md#adding-a-new-ruby-version), this should be all that's needed to add support for today's new Ruby releases? But this is my first time contributing to this repo, so please, let me know if I've missed something!


eregon:
The dist/index.js needs to also be updated, like https://github.com/ruby/setup-ruby/pull/234 does it.
Since I can only merge one of the PRs I'll merge that one since it already has all the changes.
Thank you for the PR though.

Issue #231, opened 2021-11-23
Update @actions/cache version to 1.0.8


aparna-ravindra:
**Description:**
The action depends on [`@actions/cache`](https://github.com/ruby/setup-ruby/blob/f20f1eae726df008313d2e0d78c5e602562a1bcf/package.json#L25). The latest version `1.0.8` is now available and it supports upto 10GB cache upload. 

**Justification:**
Support 10GB cache upload using the latest version `1.0.8` of [`@actions/cache` ](https://www.npmjs.com/package/@actions/cache)

PR #236, opened 2021-11-30
Updating actions/cache version 1.0.8

aparna-ravindra:
The new version of `@actions/cache` allows for caching dependencies upto 10GB. 

Closes Issue: https://github.com/ruby/setup-ruby/issues/231


eregon (approved):
Thanks!

PR #253, opened 2021-12-31
Support Ruby 3.1.0 for Windows

koic:
Follow up to https://github.com/oneclick/rubyinstaller2/issues/255#issuecomment-1003266597.

This PR adds Ruby 3.1.0 to windows-version.js.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,14 +23,12 @@ jobs:
         - { os: windows-2019, ruby: mswin }
         exclude:
         - { os: windows-2016, ruby: 1.9 }
-        - { os: windows-2016, ruby: 3.1 }
         - { os: windows-2016, ruby: debug }
         - { os: windows-2016, ruby: truffleruby }
         - { os: windows-2016, ruby: truffleruby-head }
         - { os: windows-2016, ruby: truffleruby+graalvm }
         - { os: windows-2016, ruby: truffleruby+graalvm-head }
         - { os: windows-2019, ruby: 1.9 }
-        - { os: windows-2019, ruby: 3.1 }
         - { os: windows-2019, ruby: debug }
         - { os: windows-2019, ruby: truffleruby }
         - { os: windows-2019, ruby: truffleruby-head }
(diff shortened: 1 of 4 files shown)


eregon (approved):
Thanks!

(I removed the extra newline)

eregon:
Unfortunately it fails:
https://github.com/ruby/setup-ruby/runs/4690740303?check_suite_focus=true
https://github.com/ruby/setup-ruby/runs/4690739904?check_suite_focus=true
```
  C:\Windows\system32\cmd.exe /D /S /C "C:\hostedtoolcache\windows\Ruby\3.1.0\x64\bin\gem.cmd install bundler -v "~> 2""
  ERROR:  Error installing bundler:
  	"bundle" from bundler conflicts with C:/hostedtoolcache/windows/Ruby/3.1.0/x64/bin/bundle
```
@larskanis Does RubyInstaller change bin/bundle? That causes issues like this, I'm not sure how to fix.
cc @MSP-Greg

larskanis:
> @larskanis Does RubyInstaller change bin/bundle? That can cause issues like this, I'm not sure how to fix.

No, there's no modification from the RubyInstaller side. I can reproduce this issue locally. Seems to be a problem in rubygems. I'll check this.

For the time being, the workaround is to call
```sh
gem install bundler --force
```

eregon:
@larskanis Thanks for the quick reply. @MSP-Greg found a workaround by updating to latest RubyGems (https://github.com/ruby/setup-ruby/pull/224#discussion_r777453098), so maybe it's a RubyGems/Ruby bug as in 3.1.0 ship with a bin/bundler different than the one RubyGems would generate?

eregon:
I'll try `--force` to unblock the situation, let's see if that works.
#224 likely needs more time (notably I need to review in depth)

MSP-Greg:
Just to be clear, Windows Ruby 3.1.0, head and ucrt are compiled with the MSYS2 UCRT64 tool chain, which is not installed on either windows-2019 or windows-2022.  Extension gems or c source may compile on windows-2019, which currently has MINGW64 tools. 

Note that they may or may not pass their CI, but mixing the MINGW64 & UCRT64 tools chains is a bad idea and can cause very odd problems.  MINGW64 links to a set of Windows 'legacy' dll files, UCRT64 links to a newer set.

eregon (approved):

eregon:
This PR passes all the tests now, I think it's worth merging so people can try Ruby 3.1 on Windows.
I agree we need the fixes from #224 but I also think it will need some more time to be ready to be merged.
It's the same situation with ruby-head on Windows, and so far nobody complained about it.

koic:
Thank you @eregon and @MSP-Greg!

PR #264, opened 2022-01-16
Gemfile - add Rack and RDoc for testing - failures with Ruby 2.4 & 2.5

MSP-Greg:
This seems to intermittently fail.  The RubyGems that ships with Ruby 2.4 and 2.5 is not compatible with the Psych version that can be installed.

RDoc has a dependency on Psych.  Rack is added just so `bundle install --jobs 4` probably needs to run a job after Psych is installed.

This is the weird failure that I spoke of, which is why I had `gem update --system` in the PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,4 +3,7 @@ source 'https://rubygems.org'
 
 gem "rake"
 gem "path"
+gem "rdoc"
 gem "json", "2.2.0"
+gem "rack"
+


MSP-Greg:
I believe this problem is about dependencies within the std-lib/bundled gems.  Psych runs CI with its tests running against core Ruby.  So, it does work with core Ruby 2.4 & 2.5.  Problem is, it doesn't work with the RubyGems in 2.4 & 2.5.

How to fix?

MSP-Greg:
@deivid-rodriguez @eregon

Deivid, sorry for the ping.  Noticing some odd things with Actions.

Let's say I have a Gemfile which installs Psych, which currently is compatible with Ruby 2.4 and later.  There is no Gemfile.lock file, and all the gems are installed in `vendor/bundle`, which is set with `bundle config`.  Bundler is updated, but not RubyGems.

With Ruby 2.5 (and maybe sometimes 2.4, not sure), running `bundle install` seems to generate errors like:
```
  --- ERROR REPORT TEMPLATE -------------------------------------------------------
  
  ArgumentError: wrong number of arguments (given 4, expected 1)
    /home/runner/work/ruby-setup-ruby/ruby-setup-ruby/vendor/bundle/ruby/2.5.0/gems/psych-4.0.3/lib/psych.rb:323:in `safe_load'
    /opt/hostedtoolcache/Ruby/2.5.9/x64/lib/ruby/2.5.0/rubygems/safe_yaml.rb:31:in `safe_load'
    /opt/hostedtoolcache/Ruby/2.5.9/x64/lib/ruby/2.5.0/rubygems/package.rb:496:in `block (2 levels) in read_checksums'
```

This is the situation when Action's restores a Bundle cache.  I haven't tried it locally yet, wondering if you think this may be the issue?  It doesn't seem to happen if `vendor/bundle` isn't populated with gems, similar to the cache not being used in Actions...

dentarg:
Here's example runs of the issue @MSP-Greg talks about above

* `bundler-cache: false` Ruby 2.5 works: https://github.com/dentarg/puma/runs/4834198864?check_suite_focus=true#step:9:71
* `bundler-cache: true` Ruby 2.5 fails: https://github.com/dentarg/puma/runs/4837582771?check_suite_focus=true#step:4:2054

deivid-rodriguez:
I did't have a close look, but the issue is indeed RubyGems depending on Psych internally. This has been partly address in recent RubyGems versions by supporting both Psych 3 and 4, so the solution is to update RubyGems when it's not modern enough to include that fix. Since `ruby/setup-ruby` has accepted to allow updating RubyGems, that seems enough to address this issue, no? Of course, someone has to implement it, but in the mean time I guess you can give up on bundler-cache and update RubyGems manually.

What I mean is that yes, this issue is intermittent depending on whether Bundler ends up hitting the Psych 4 interface old RubyGems is not compatible with. But it doesn't seem worth to me digging further given that upgrading RubyGems fixes the issue in general.

MSP-Greg:
1. I have repo'd this locally.  Psych needs to be in the Gemfile, and I believe `bundle install` needs the `--jobs` option.
2. Found an example of this happening about 3 weeks ago, see https://github.com/ruby/rdoc/runs/4646871718?check_suite_focus=true#step:3:57.  Note that the CI was updating RubyGems for Ruby 2.5 and 2.6 after setup-ruby.

Now, this is breaking CI for repos, but it's nothing to do with the action.  Since the action allows `bundle install` within it, what to do?

One option is to add an input for RubyGems.  I'm not sure what I think of that, or the possible support problems with its interaction with a Bundler input.

Another option might be to add bundler options of something like `gem-update-sys-fix` and `gem-update-sys-all`.  The first would run `gem update --system` for Rubies 2.3 thru 2.6, the second would do so for 2.3 thru 3.1.  Obviously, this would install Bundler 2 also.

Thoughts?

deivid-rodriguez:
If I understood correctly, @eregon accepted a `rubygems` input, it only needs someone to implement it.

Issue #228, opened 2021-11-08
[Feature Request] Allow changing Rubygems version


ashmaroli:
Currently, if I were to run a job using Ruby 2.4 and Rubygems 2.x, I'm locked to `rubygems-2.6.14.4`.
If there's a gem that *requires* Rubygems 2.7 (for security reasons), `bundle install` fails, forcing me to abandon the bundle-cache provided by this action.

I would therefore like to have a way to update to `rubygems-2.7.11`. Under the hood, it would be a system call to
```
gem update --system --version="~> 2.7"
```

dentarg:
Maybe it makes sense to ensure at least RubyGems 2.7.11 in `setup-ruby`? RubyGems 2.7.11 seems to work all the way back to [Ruby 1.9](https://github.com/spinels/rack-ssl-enforcer/commit/9124de78d011acd5676b77112b33e93d7f3576fc).

eregon:
I dislike updating RubyGems, for many reasons, https://github.com/ruby/setup-ruby/pull/224#discussion_r777453098 and https://eregon.me/blog/2020/01/13/a-migration-path-to-bundler2.html#the-rubygems-requirement-of-bundler-2.
So I don't want to update RubyGems automatically in setup-ruby.

But of course if people want to update to a given RubyGems version, then that's fine as it's their explicit choice.
Will be solved when the caching is a separate action.

Jack12816:
Maybe it could be turned on as a boolean input? The default could be off.

Currently, I'm getting this issue for a vanilla 2.5 build:

```
    steps:
      - uses: actions/checkout@v2

      - name: Install the correct Ruby version
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
          bundler-cache: true      
```

```
Run ruby/setup-ruby@v1
  with:
    ruby-version: 2.5
    bundler-cache: true
    bundler: default
    working-directory: .
    cache-version: 0
  env:
    BUNDLE_GEMFILE: gemfiles/rails_5.2.gemfile
Modifying PATH
  Entries added to PATH to use selected Ruby:
    /opt/hostedtoolcache/Ruby/2.5.9/x64/bin
Installing Bundler
  /opt/hostedtoolcache/Ruby/2.5.9/x64/bin/gem install bundler -v ~> 2
  Successfully installed bundler-2.3.4
  1 gem installed
  Took   2.67 seconds
bundle install
[...]

PR #271, opened 2022-01-21
Add a new input to optionally update RubyGems

deivid-rodriguez:
Implements #228.

I guess TruffleRuby needs some special handling. Also, I'd like to ask what's the recommended way to try out my changes when working on a PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,10 @@ setting the `rubygems` input. See
 
 See [action.yml](action.yml) for more details about the `rubygems` input.
 
+If you're running into `ArgumentError: wrong number of arguments (given 4,
+expected 1)` errors with a stacktrace including Psych and RubyGems entries, you
+should be able to fix them by setting `rubygems: 3.0.0` or higher.
+
 ### Bundler
 
 By default, if there is a `Gemfile.lock` file (or `$BUNDLE_GEMFILE.lock` or `gems.locked`) with a `BUNDLED WITH` section,
--- a/README.md
+++ b/README.md
@@ -139,6 +139,14 @@ and the [condition and expression syntax](https://help.github.com/en/actions/ref
 The `working-directory` input can be set to resolve `.ruby-version`, `.tool-versions` and `Gemfile.lock`
 if they are not at the root of the repository, see [action.yml](action.yml) for details.
 
+## RubyGems
+
+By default, the default RubyGems version that comes with each Ruby is used.
+However, users can optionally customize the RubyGems version that they want by
+setting the `rubygems` input. See
+
+See [action.yml](action.yml) for more details about the `rubygems` input.
+
 ### Bundler
 
 By default, if there is a `Gemfile.lock` file (or `$BUNDLE_GEMFILE.lock` or `gems.locked`) with a `BUNDLED WITH` section,
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ if they are not at the root of the repository, see [action.yml](action.yml) for 
 
 By default, the default RubyGems version that comes with each Ruby is used.
 However, users can optionally customize the RubyGems version that they want by
-setting the `rubygems` input. See
+setting the `rubygems` input.
 
 See [action.yml](action.yml) for more details about the `rubygems` input.
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -123,16 +123,27 @@ jobs:
         rubygems: latest
     - run: ruby -e "exit(Gem.rubygems_version > Gem::Version.new('3.0.3'))"
 
-  testFixedRubygemsVersion:
-    name: "Test rubygems input set to a fixed version upgrades RubyGems to that versoin"
+  testFixedRubygemsVersionUpgrades:
+    name: "Test rubygems input set to a fixed version upgrades RubyGems to that version if the default is older"
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
     - uses: ./
       with:
         ruby-version: 2.6
-        rubygems: 3.3.5
-    - run: gem --version | grep -F "3.3.5"
+        rubygems: 3.2.3
+    - run: gem --version | grep -F "3.2.3"
+
+  testFixedRubygemsVersionNoop:
+    name: "Test rubygems input set to a fixed version noops if the default is newer"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./
+      with:
+        ruby-version: 3.1
+        rubygems: 3.2.3
+    - run: gem --version | grep -F "3.3.3"
 
   testExactBundlerVersion:
     name: "Test with an exact Bundler version"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,6 +112,28 @@ jobs:
       if: startsWith(matrix.os, 'windows') && startsWith(matrix.ruby, 'jruby')
       run: gem install sassc
 
+  testLatestRubygemsVersion:
+    name: "Test rubygems input set to latest upgrades the default RubyGems version"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./
+      with:
+        ruby-version: 2.6
+        rubygems: latest
+    - run: ruby -e "exit(Gem.rubygems_version > Gem::Version.new('3.0.3'))"
+
+  testFixedRubygemsVersion:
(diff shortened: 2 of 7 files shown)


deivid-rodriguez:
I see that CI already exercises everything, awesome!

eregon:
Thank you for the PR!

Could you add a single CI job around https://github.com/ruby/setup-ruby/blob/180e1d1f1c5eadbfbab3dab6fb79ab0c07e3cecc/.github/workflows/test.yml#L115-L168 to test updating to latest, so we make sure it works and keeps working?
Maybe also for a specific version.

Since the default remains `default` then the I think big matrix should still use `default`.

And could you add a new `RubyGems` short section section in the README about this new input?

eregon:
For TruffleRuby it will just ignore the update, which is probably fine for now, so I think best to treat it like any other Ruby in this action.

eregon:
What's your opinion regarding the psych 4 + old Ruby/RubyGems issue? (https://github.com/ruby/setup-ruby/pull/264)

I think updating to a given RubyGems version wouldn't work if people have a matrix of both old and new Rubies (it would actually downgrade RubyGems on the newer Rubies).
Maybe a special version `psych4` or so would make sense? (feels hacky to me)
Or maybe never downgrade and always skip updating if the installed RubyGems is more recent than the given version (sounds nice to me).
(Updating to latest RubyGems would probably work for all but is adding more moving parts in CI)

deivid-rodriguez:
Thanks, I will make the requested updates 👍.

Regarding the Psych + old RubyGems issue, a custom input value feels too hacky, yeah. I saw you added some custom code sometimes to workaround issues with bundler on different rubies (for example, installing Bundler 1 on JRuby 9.1 because it has issues with Bundler 2). I guess we could similarly automagically update RubyGems to 3.0.3 for example (the same shipped with Ruby 2.6) on older rubies and note that in the "rubygems: default" documentation. Not sure if the issue is spread enough, as far as I understand for the issue to happen, the following conditions must be met:

* Running Ruby 2.4 or 2.5, because older Rubies don't support Psych 4, and newer Rubies come with a recent enough Rubygems.
* `bundler-cache: true` (I guess this is common because it's the default?).
* Psych 4 being in cache, and thus being restored before `bundle install` is run.

eregon:
Yeah, we could e.g. look at the lockfile to see if there is Pysch 4 in there, but that feels hacky.
I think we don't try to address this directly, it's probably not frequent enough to warrant it, and users should probably understand Psych 4 + Ruby 2.4-2.5 is a problematic combination anyway (not just for CI but also for anyone running those locally).

> bundler-cache: true (I guess this is common because it's the default?).

It's not default but it's fairly common.

eregon:
@deivid-rodriguez  What do you think of this:
> Or maybe never downgrade and always skip updating if the installed RubyGems is more recent than the given version (sounds nice to me).

I think that seems generally useful, unless there might be a valid reason to downgrade RubyGems lower than the version with Ruby? (I guess not)

deivid-rodriguez:
> I think that seems generally useful, unless there might be a valid reason to downgrade RubyGems lower than the version with Ruby? (I guess not)

Oh, we automatically prevent that. If you try to downgrade RubyGems to a version older than the one that comes with each Ruby, it raises an error. Of course, you need to start from a RubyGems version that has this feature, and it's fairly recent.

deivid-rodriguez:
I think downgrading can be useful sometimes, for example, to workaround some regression. But we forbid downgrading past the default version that comes with Ruby because we never test that kind of combination.

eregon:
OK, then we should probably check `gem --version` in this action if the rubygems input is set to a given version as we wouldn't want to fail the workload in such a case.
For example in Puma's CI where the psych 4 issue happens they'd want something like:
```yml
jobs:
  test:
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-latest, macos-latest]
        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1]
    runs-on: ${{ matrix.os }}
    steps:
    - uses: actions/checkout@v2
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: ${{ matrix.ruby }}
        rubygems: MINIMUM_RUBYGEMS_VERSION_WHICH_SUPPORTS_PSYCH4
        bundler-cache: true
```
and for Ruby >= 2.6 it should just no try to update RubyGems as that would error (or downgrade).

deivid-rodriguez:
Oh, you're right. Yeah, that's probably best, if a specific version is given, interpret and document it as a minimum RubyGems version for the matrix, and do nothing if the default one for a particular Ruby is newer.

> BTW, which RubyGems version has Psych 4 support?

I think it's 3.0.0.beta2: https://github.com/rubygems/rubygems/commit/b7995ca4ea8452b4511f4afa7eddd5374fca21fc.

deivid-rodriguez:
I was thinking of recommending using either 3.0.3 (the version of RubyGems that comes with the next Ruby, 2.6), or 3.0.9 (the latest patch level on that series). I think 3.0.3 should be fine.

deivid-rodriguez:
I added the requested changes. CI is having some issues on Windows, but it's not this PR's fault.

deivid-rodriguez:
Also added a note recommending `rubygems: 3.0.0` for the Psych + RubyGems issue.

eregon:
cc @MSP-Greg for the windows failures (mingw & ucrt):
https://github.com/ruby/setup-ruby/runs/4905624491?check_suite_focus=true
https://github.com/ruby/setup-ruby/runs/4905624491?check_suite_focus=true
https://github.com/ruby/setup-ruby/runs/4905624590?check_suite_focus=true
https://github.com/ruby/setup-ruby/pull/258 might fix it but then I'd want to first do that only for mingw & ucrt builds.

eregon:
This is awesome, thank you!

Some things I noticed in https://github.com/ruby/setup-ruby/runs/4905624850?check_suite_focus=true
* The output of updating RubyGems takes > 1000 lines, I can imagine that to be impractical when looking at the full log. Is there a way to avoid printing the changelog of all versions in between?
* `gem --version` gives `3.0.3.1` but then just below it says `Default RubyGems version is 3.0.3`. Is there a way to have the semver package handle versions with not exactly 3 components?

MSP-Greg:
Can you rerun the CI?  Not sure what happened, last night something went haywire with MSYS2, and I've never seen it before.

Since I've never seen it before, I didn't have any tests to fail the 'builds', so they replaced the working release assets.

That cascaded into the ruby-loco builds.  I fixed things late last night (early this morning), and I'm adding tests now to both.

README.md
```diff
@@ -139,6 +139,18 @@ and the [condition and expression syntax](https://help.github.com/en/actions/ref
 The `working-directory` input can be set to resolve `.ruby-version`, `.tool-versions` and `Gemfile.lock`
 if they are not at the root of the repository, see [action.yml](action.yml) for details.
 
+## RubyGems
+
+By default, the default RubyGems version that comes with each Ruby is used.
+However, users can optionally customize the RubyGems version that they want by
+setting the `rubygems` input. See
```
eregon, line 146:
```suggestion
setting the `rubygems` input.
```

eregon:
> * `gem --version` gives `3.0.3.1` but then just below it says `Default RubyGems version is 3.0.3`. Is there a way to have the semver package handle versions with not exactly 3 components?

It seems the `semver` package doesn't handle this. Maybe it doesn't matter too much in practice, WDYT?

eregon:
@MSP-Greg The rerun is all green indeed, thanks for the reply.

deivid-rodriguez:
Thank you @eregon.

Regarding your first question, we have `gem update --system --silent`, although we only fixed that recently to not show the changelog of what was installed. Should I use that instead?

Regarding `3.0.3.1` vs `3.0.3`, I noticed that too and I'm not 100% sure what's going on there. I can say that 3.0.3.1 is not released to rubygems.org, and in general I don't think we release 4 digit versions to rubygems.org. At least `gem update --system` does not support 4 digit versions. My best guess is that ruby-core used to ship released versions of RubyGems with some custom patches on top, adding an extra digit to the version every time they did that. I think we can pretend this last digit does not exist.

eregon:
I tried `--silent` locally and it is indeed very silent: there is no output at all.
The rest of the output actually seems useful, notably to show the installer Bundler and RubyGems versions.
Filtering the output manually seems rather messy, so I guess the best is to just keep the verbose output.

eregon (approved):

deivid-rodriguez:
Yes, I agree with you, I think we should have some intermediate option between `--silent` and the default.

Thanks for merging!

Issue #221, opened 2021-09-25
Remove binaries for Ubuntu 16.04


eregon:
Ubuntu 16.04 is no longer supported on GitHub Actions: https://github.com/actions/virtual-environments/issues/3287
So we can remove 16.04 binaries from https://github.com/ruby/ruby-builder/releases and remove any 16.04-specific code.
Will wait at least until October 18, 2021 as Azure DevOps might rely on those.

I guess self-hosted runners could rely on those if the self-hosted runner uses Ubuntu 16.04.
I think we should drop them, as new Rubies can't be built on 16.04 in GitHub Actions anyway, and supporting a platform we can't test is not reasonable.

eregon:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources says:
> Deprecated and limited to existing customers only. Migrate to Ubuntu 20.04.

Not sure what that means.

hsbt:
I'm +1 to remove it after Oct 18, 2021. In fact, I removed Ubuntu 16.04 from https://rubyci.org/. Ruby core team are not guarantee to support the all versions of Ruby with Ubuntu 16.04.

eregon:
Done, 375 builds -> 299 builds for https://github.com/ruby/ruby-builder/releases/tag/toolcache

PR #298, opened 2022-03-14
Remove ubuntu-16.04 from the list of supported Ubuntu versions.

cliffano:
ubuntu-16.04 is no longer supported as noted on https://github.com/ruby/setup-ruby/issues/221 .
Readme should no longer list ubuntu-16.04 on the supported Ubuntu versions note.


eregon (approved):
Thanks

PR #302, opened 2022-03-31
Use actions/checkout@v3

kaiquekandykoga:


eregon:
What's the advantage of v3? I don't find much on https://github.com/actions/checkout.
It seems a node version update, why should this action care about that? (this action uses node12 BTW)
And should the documentation be updated too?

kaiquekandykoga:
I was just trying to keep it up to date. I changed only the workflows, because I have tested only the workflow in my fork and not the examples in the documentation (README.md).

I am closing it.

eregon:
Thank you for the PR, we may end up merging it.
I really question the need of a new major version of `actions/checkout` just to bump the node version though, that seems needlessly annoying for the majority of workflows (https://github.com/actions/checkout/pull/708#issuecomment-1084469437).

PR #306, opened 2022-04-05
Fix `bundler:` with a minor version installing the wrong version

### LINKED ISSUE #304: Can't install latest bundler for a specific minor version (e.g. 2.2.x)
Let's say I want to install the latest bundler 2.2.x.

- If I pass `bundler: 2.2`, then it installs `2.3.x` because it generates the command `gem install bundler -v ~> 2.2`
- if I pass `bundler: 2.2.0`, then it installs `2.2.0`, and not the latest, because it generates the command `gem install bundler -v 2.2.0`.
- If I pass `bundler: 2.2.x` that is not valid input into the gem install command.
- I've also tried mangling the input in various ways to bypass the [regex](https://github.com/ruby/setup-ruby/blob/bd94d6a504586da892a5753afdd1480096ed30df/bundler.js#L98) that gates the `~>`, but those are detected as [invalid input](https://github.com/ruby/setup-ruby/blob/bd94d6a504586da892a5753afdd1480096ed30df/bundler.js#L68).

If there is no workaround, then this is an enhancement request to perhaps allow something like `bundler: 2.2.x`, which, if detected, would translate to `gem install bundler -v ~> 2.2.0`.

Example:

```yaml
    - name: Set up Ruby
      uses: ruby/setup-ruby@v1
      with:
        ruby-version: ${{ matrix.ruby-version }}
        bundler: 2.2.x
        bundler-cache: true
```

Fryguy:
If passed `bundler: 2.2`, setup-ruby would install bundler 2.3, because
it would generate a constraint that looks like `~> 2.2`. This
unexpectedly installs the latest of the 2.x series, but the expectation
is that it would install the latest of the 2.2.x series.

Fixes #304


Fryguy:
This PR also has the dist.js updated.  I wasn't sure what your PR pattern is in this repo, so if this is wrong, let me know and I can remove that.

eregon (approved):
Thank you for the PR, this looks great.

eregon:
The CI seems to fail, see https://github.com/ruby/setup-ruby/runs/5836332794?check_suite_focus=true
Could you try to figure it out? A separate step before to run just `bundle --version` would likely be helpful.

.github/workflows/test.yml
```diff
@@ -145,6 +145,28 @@ jobs:
...
+  testMajorBundlerVersion:
+    name: "Test with a major Bundler version"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./
+      with:
+        ruby-version: 2.6
+        bundler: 2
+    - run: bundle --version | grep "Bundler version 2\.\d+.\d+"
```
eregon: I think `grep` without any flag doesn't understand `\d+`.
It' seems to work with `grep -P`:
```
$ echo 2.2.32 | grep -P "2\.2\.\d+"
2.2.32
```

.github/workflows/test.yml
```diff
@@ -154,7 +154,7 @@ jobs:
       with:
         ruby-version: 2.6
         bundler: 2
-    - run: bundle --version | grep "Bundler version 2\.\d+.\d+"
+    - run: bundle --version | grep "Bundler version 2\.\d\+.\d\+"
```
eregon, line 157: This pattern doesn't work for me locally, so it seems unlikely to work on GHA:
`echo 2.2.32 | grep "2\.\d\+.\d\+"` => exit 1, no output

Fryguy:
Ok, updated with the `-P`.  Frustratingly, the original PR worked in [act](https://github.com/nektos/act).

PR #315, opened 2022-04-17
Windows - updates for mswin builds

MSP-Greg:
Updates for mswin builds, no 'bugs' involved.  Commit info:

1. '[package.json and yarn.lock](https://github.com/ruby/setup-ruby/commit/55996aeeb97e7655ea5afacd23e44b6600db10d3)' - 'housekeeping'
2. '[Windows mswin - update addVCVARSEnv](https://github.com/ruby/setup-ruby/commit/c3a022e6a96d4f8c0ae1cf414623a5f4c19e57ec)' - Currently the 'vcvars.bat' file's path was found via a switch/select/case statement that hard-coded the Actions windows image string.  A PowerShell module ([vssetup.powershell](https://github.com/microsoft/vssetup.powershell)) is available for finding information about Visual Studio installs, and it is installed on all Actions Windows runners.  Use the module so hard-coded image strings aren't needed.
3. '[Windows mswin - use ucrt64 tools instead of mingw64](https://github.com/ruby/setup-ruby/commit/58e1d35f716c6bd65bb87f52bc3b4d7b956e8b22)' - Many repos need bash commands, and also use build utilities like bison or ragel.  Currently, mswin builds are using the mingw64 tools, switch to the ucrt64 tools.
4. '[Actions - add windows-2022 mswin job](https://github.com/ruby/setup-ruby/commit/31be906aaecd36952ac494328a126f7e2c664495)'

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -199,24 +199,17 @@ async function setupMSWin() {
  *   adds a convenience VCVARS environment variable
  *   this assumes a single Visual Studio version being available in the Windows images */
 export function addVCVARSEnv() {
-  let vcVars = ''
-  switch (virtualEnv) {
-    case 'windows-2016':
-      vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
-      break
-    case 'windows-2019':
-      vcVars = '"C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
-      break
-    case 'windows-2022':
-      vcVars = '"C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Auxiliary\\Build\\vcvars64.bat"'
-      break
-    default:
-      throw new Error(`Unknown Windows Image: ${virtualEnv}`)
+  let cmd = `pwsh.exe -command "Get-VSSetupInstance -All | Select-VSSetupInstance -Latest | Select-Object -ExpandProperty InstallationPath"`
+  let vcVars = `${cp.execSync(cmd).toString().trim()}\\VC\\Auxiliary\\Build\\vcvars64.bat`
+
+  if (!fs.existsSync(vcVars)) {
+    throw new Error(`Missing vcVars file: ${vcVars}`)
   }
   core.exportVariable('VCVARS', vcVars)
 
+  cmd = `cmd.exe /c ""${vcVars}" && set"`
+
   let newEnv = new Map()
-  let cmd = `cmd.exe /c "${vcVars} && set"`
   let newSet = cp.execSync(cmd).toString().trim().split(/\r?\n/)
   newSet = newSet.filter(line => /\S=\S/.test(line))
   newSet.forEach(s => {
--- a/windows.js
+++ b/windows.js
@@ -199,7 +199,7 @@ async function setupMSWin() {
  *   adds a convenience VCVARS environment variable
  *   this assumes a single Visual Studio version being available in the Windows images */
 export function addVCVARSEnv() {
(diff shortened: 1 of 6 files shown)


MSP-Greg:
Original PR used a PowerShell module known as [vssetup.powershell](https://github.com/microsoft/vssetup.powershell), which contains quite a bit of functionality.  But, running it took anywhere from 14 to 20 sec in one run.  Too long.

Windows images also contain a utility named [vsWhere](https://github.com/Microsoft/vswhere), which is a compiled exe.  It takes around a second to run.  Last commit uses it...

windows.js
```diff
@@ -199,24 +199,17 @@ async function setupMSWin() {
...
-      throw new Error(`Unknown Windows Image: ${virtualEnv}`)
+  let cmd = 'vswhere -latest -property installationPath'
+  let vcVars = `${cp.execSync(cmd).toString().trim()}\\VC\\Auxiliary\\Build\\vcvars64.bat`
+
+  if (!fs.existsSync(vcVars)) {
+    throw new Error(`Missing vcVars file: ${vcVars}`)
   }
   core.exportVariable('VCVARS', vcVars)
 
+  cmd = `cmd.exe /c ""${vcVars}" && set"`
```
eregon, line 210: Are the double `"` intended here?

eregon (approved):
Looks great, thank you and sorry for the review delay

windows.js
MSP-Greg: Yes.  There are two pairs, an outer and an inner.  Kind of a pita when Actions uses PowerShell Core for the default script processor, but almost all languages (like Ruby & node.js ) shell out using cmd script.  The two have different styles for handling embedded quotes in commands...

MSP-Greg:
> sorry for the review delay

No problem.  OSS is just a part of our lives...

Do you have any opinion about whether we/I should build the mswin Ruby with OpenSSL 3.0?  Microsoft's vcpkg 'system' has updated to 3.0 from 1.1.  I've also got all the packages needed to build mswin in the [MSP-Greg/setup-msys2-gcc](https://github.com/MSP-Greg/setup-msys2-gcc/releases/tag/msys2-gcc-pkgs) release.  Hence, it could be installed by setup-ruby just like the mingw and ucrt archive files.

I think I'm aware of most of the repos it would affect, and most already work with OpenSSL 3.0...

eregon:
> Do you have any opinion about whether we/I should build the mswin Ruby with OpenSSL 3.0?

I think all GitHub-hosted runners currently use openssl 1.1, not 3, so probably best to stick to 1.1 while it's the case (unless that's problematic for mswin).

MSP-Greg:
GitHub has announced that Ubuntu 22.04 will be available as a preview image some soon.  I believe it uses OpenSSL 3.

I'm running it locally, and building/using Ruby master, even added gcc 12.

Also, the current release of Microsoft's vcpkg system [2022.04.12](https://github.com/microsoft/vcpkg/blob/2022.04.12/ports/openssl/vcpkg.json) builds OpenSSL 3, so sometime soon the Actions Windows runners would build a mswin package with 3.

MSP-Greg:
One thing I haven't mentioned.  mswin builds use a different set of packages than either mingw or ucrt.  In the past, it was kind of a mishmash of sources.  Now, [microsoft/vcpkg](https://github.com/microsoft/vcpkg) has all the packages needed to build mswin (the bison and ragel utilities from MSYS2 work with mswin).

Just last month vcpkg updated OpenSSL from 1.1.1 to 3.0 ([diff](https://github.com/microsoft/vcpkg/commit/407bc79b391a8e30866c95294babca5b111f6006#diff-cb9263a1f84e21b83a97ea8a1bcef649208f2cedafbeea9a50002cdacfca3ed9R3)).  So, a new installation of vcpkg builds OpenSSL 3.0.2.  Packages from vcpkg can be 'exported', so I've also got an archive file of all the packages needed to build mswin.

eregon:
Sounds good. Yeah then it sound like using vcpkg packages is the best for `mswin`.

PR #316, opened 2022-04-20
Clarify reference to a known issue now fixed

deivid-rodriguez:
The culprit ended up being https://github.com/jruby/jruby/issues/6745, which was released with JRuby 9.2.20.0.

Happy to change the wording to make this clear instead of removing the reference.


eregon:
Maybe we should keep it and mention it like `JRuby < 9.2.20.0 ...`?
Specifically I'm happy if people don't file an issue again about it, and know JRuby on Windows also has/had some bugs on Windows and not immediately consider it a setup-ruby bug.

deivid-rodriguez:
In my opinion, explicitly mentioning every third party issue that has ever affected this action seems too much, that's why I went with removing the mention. But as I said, I'm happy to update it as you see fit!

deivid-rodriguez:
I gave it another try that I suspect you will like better.

eregon (approved):
Thanks!

PR #330, opened 2022-05-08
MSP-Greg/setup-msys2-gcc -> ruby/setup-msys2-gcc, update packages

MSP-Greg:


windows.js
```diff
@@ -14,11 +14,12 @@ const rubyInstallerVersions = require('./windows-versions.json')
 const drive = common.drive
 
 const msys2BasePath = 'C:\\msys64'
+const msys2GCCReleaseURI  = 'https://github.com/ruby/setup-msys2-gcc/releases/download'
 
-// needed for 2.0-2.3, and mswin, cert file used by Git for Windows
+// needed for Ruby 2.0 thru 2.3, and mswin, cert file used by Git for Windows
```
eregon, line 19: I'm not a fan of `2.0 thru 2.3` (looks like badly spelled C code), it could be `through` but that's long and not easy to read.
Can you use use `for Ruby 2.0-2.3`, same below?

ahayworth:
Hi all - This is causing all of our Windows ruby builds to fail. I hate to ask fellow open source maintainers to jump on a problem ... but is it possible to expedite the review of this PR? It would mean a lot to us if you're able to do so. ❤️ ❤️

MSP-Greg:
@ahayworth

My mistake, MSP-Greg/setup-msys2-gcc generates archive files for ruby/setup-ruby.  It was recently transferred to the ruby org GitHub account.  That has nothing to do with the current issue.

What happened this morning, is that a workflow that saves three release assets succeeded with two, but the third threw a '502 Bad Gateway' error.  In that repo and another, I've run several thousand jobs and never had that error.

I'll update the code to retry, but at present, I can't fix the error myself.

ahayworth:
@MSP-Greg no worries in the slightest, chasing down build problems can be a nightmare. Having spent the past few days fighting with Actions myself (for completely unrelated problems), I definitely understand.

Thanks for getting back to me though, it's much appreciated. We'll be alright until things are sorted here. 😄

ahayworth:
> @ahayworth Fixed. Again, sorry for the failed jobs...

You're amazing, thank you!

MSP-Greg:
Rebased, and there was a deprecation warning being logged:
```
Extracting  msys2 build tools
  (node:2352) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
  (Use `node --trace-deprecation ...` to show where the warning was created)
  Took  10.60 seconds
```

eregon (approved):
Thanks!

PR #332, opened 2022-05-10
Update Node.js from v12 to v16

yykamei:
According to the release policy of Node.js,
Node.js v12 reached EOL (end-of-life).

https://nodejs.org/en/about/releases/

I think ruby/setup-ruby is ready to update Node.js to v16
because GitHub Actions runners support v16.

https://github.com/actions/runner/issues/772

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-nodejs 12.0.0
+nodejs 16.0.0
 ruby 2.7.0
--- a/action.yml
+++ b/action.yml
@@ -32,5 +32,5 @@ outputs:
   ruby-prefix:
     description: 'The prefix of the installed ruby'
 runs:
-  using: 'node12'
+  using: 'node16'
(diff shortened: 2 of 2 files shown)


eregon (approved):
Thanks!

chenrui333:
I am a bit surprised that this did not come as a major release (there are incompatibility with this change) Any thoughts?

ShockwaveNN:
> Note that the [Ubuntu-22.04](https://github.com/actions/virtual-environments/blob/2e10aa4cfd64f665e133fe292d0b7b4bbd0e00e1/images/linux/Ubuntu2204-Readme.md#nodejs) image only has node 14 & 16 installed.

I've take a look and Ubuntu 18.04, Ubuntu 20.04 and Ubuntu 22.04, all macOS and latest Windows images have nodejs v16 installed by default according to README

So not sure about that incompatibility @chenrui333 talking about?

Issue #337, opened 2022-05-16
Any idea when macOS-12 will be supported?


duro:
Using `runs-on: macos-12` I currently get `Error: Unsupported platform macos-12.0`

Curious if there is a timeline on when this action will support that runner?

matbouil:
Hi ! I encountered this error too ... However it was still working with macos-12.0 three days ago.

duro:
I went ahead and created a pull request to simply add macos-12 to the list of allowed runners. My testing has gone flawlessly so far.

alrocha:
as a temporary fix, I set it up to the previous version which was working fine
`uses: ruby/setup-ruby@v1.106.0`

eregon:
It was working accidentally before.
It should be fine to add it to supported platforms if tests pass on this platform.

eregon:
Support added in https://github.com/ruby/setup-ruby/releases/tag/v1.108.0 (a7439cf424c9677922eee7cb72a848b9fd7fb3ff)

PR #338, opened 2022-05-16
Allow usage on macOS 12 runners

duro:
Per my question in #337 I went ahead and tested simply adding macos-12 to the supported platforms, and I had no issues.

Not sure what remaining testing may be needed, but it would appear it "just works"

### DIFF
--- a/common.js
+++ b/common.js
@@ -104,6 +104,7 @@ export const supportedPlatforms = [
   'ubuntu-22.04',
   'macos-10.15',
   'macos-11.0',
+  'macos-12.0',
   'windows-2019',
   'windows-2022',
 ]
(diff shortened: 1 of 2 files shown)


eregon:
Could you run tests for it by adapting https://github.com/ruby/setup-ruby/blob/master/.github/workflows/test.yml ?
We can't support a platform without testing it.

eregon:
It also needs an update of the README

eregon:
I'll do it myself so it's fixed faster.
I can't push to your branch so I'll redo it from scratch.

eregon:
Support added in https://github.com/ruby/setup-ruby/releases/tag/v1.108.0 (a7439cf424c9677922eee7cb72a848b9fd7fb3ff)

PR #340, opened 2022-05-17
Windows - install vcpkg archive from setup-msys2-gcc when using mswin builds

### LINKED ISSUE #331: Windows: OpenSSL 3
Currently, Windows mswin builds get most of their dependencies from [microsoft/vcpkg](https://github.com/microsoft/vcpkg).  They have updated their OpenSSL package from 1.1.1 to 3.0.x

MSYS2 (used with mingw & ucrt Ruby builds) has not yet done so, but given that vcpkg and Ubuntu 22.04 have updated to OpenSSL 3, MSYS2 will be doing so, possibly sooner rather than later.

This will require changes to this repo, along with changes to [ruby/setup-msys2-gcc](https://github.com/ruby/setup-msys2-gcc).

So, in the immediate future, I may open PR's related to switching mswin over to OpenSSL 3, and begin to work on code to determine whether to install MSYS2 OpenSSL 1.1.1 or 3.0.x packages.

Really, this issue may occur anytime one of Ruby's build dependencies has a new major release...

MSP-Greg:
Similar to current code, which install all packages/tools for building mingw and ucrt Rubies, this PR installs the set of packages from [microsoft/vcpkg](https://github.com/microsoft/vcpkg) needed to build an mswin Ruby.

This will also allow all Ruby extension gems/std-lib repos to build on mswin without using [ruby/setup-ruby-pkgs](https://github.com/ruby/setup-ruby-pkgs).

Closes #331

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -83,7 +83,9 @@ export async function install(platform, engine, version) {
     await installGCCTools(msys2Type)
   }
 
-  if (version === 'mswin') { await installVCPkg() }
+  if (version === 'mswin') {
+    await installVCPkg()
+  }
 
   const ridk = `${rubyPrefix}\\bin\\ridk.cmd`
   if (fs.existsSync(ridk)) {
--- a/windows.js
+++ b/windows.js
@@ -1,5 +1,5 @@
-// Most of this logic is from
-// https://github.com/MSP-Greg/actions-ruby/blob/master/lib/main.js
+// 7z arguments
+//   -aoa overwrite existing, -bd disable progress indicator
 
 const fs = require('fs')
 const path = require('path')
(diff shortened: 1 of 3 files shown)


windows.js
```diff
@@ -81,6 +83,8 @@ export async function install(platform, engine, version) {
     await installGCCTools(msys2Type)
   }
 
+  if (version === 'mswin') { await installVCPkg() }
```
eregon, line 86:
```suggestion
  if (version === 'mswin') {
    await installVCPkg() 
  }
```
So control flow can be seen visually without scanning the whole line

eregon (approved):

windows.js
MSP-Greg: Thanks, AFK for a bit.

PR #343, opened 2022-05-23
chore: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,12 @@
 name: Update the v1 branch when a release is published
 on:
   release:
+    permissions:
+      contents: write  # for Git to git push
     types: [published]
+permissions:
+  contents: read
+
 jobs:
   release:
     runs-on: ubuntu-latest


eregon:
Upon reading more about this, I think it is useful (https://github.com/actions/checkout/issues/485#issuecomment-1197047674), I'll merge this, thanks for the PR.
It is unfortunate that none of the links above really explains the issues/threats clearly.

eregon (approved):

eregon:
This does not work: https://github.com/ruby/setup-ruby/runs/7582388507?check_suite_focus=true

Issue #358, opened 2022-07-20
Should we use the latest Bundler release or the Bundler shipped with that Ruby by default?


eregon:
Currently it's the former but I'm thinking to change to the latter.

But it led to several issues already, including https://github.com/rubygems/rubygems/issues/5691 and previous issues where a Bundler release was broken for some aspect (I'm too lazy to search but I'd think at least 2 other issues).
To be clear, every software has bugs, I think the Bundler is doing a good job, no problem there.
But I think this is no coincidence, the Bundler shipped with a Ruby is more tested (because it's tested in that Ruby's CI rather extensively and manually during that Ruby's development) and more stable (since it doesn't change that often).

Of course, such issues in Bundler still need to be fixed and some users might need a newer Bundler than what is shipped with the Ruby.
But exposing CI users more than in local development to this kind of issues does not seem ideal.

Because in local development, one doesn't `gem install bundler` before every `bundle install`.
In fact I rarely do `gem install bundler` but tend to use the bundler shipped with that Ruby (mostly to not add another variable part in my environment/repro).

Note we already use the "shipped" Bundler for `-head` versions, but for Ruby releases we currently `gem install bundler`.
So the change would be for Ruby releases which do ship with Bundler to just use that Bundler and no longer `gem install bundler` by default. Of course one could still configure they want latest Bundler with: `bundler: latest`.

@MSP-Greg @dentarg @deivid-rodriguez Any opinion on this?

deivid-rodriguez:
I of course prefer the latest version than the default version. And I disagree the default version is more tested, sometimes the default version is only cherry-picked by ruby-core a few days days before releases. Yes, we sometimes introduce regressions, but those regressions could also happen in the default version (and it wouldn't be the first time). If you change this, you'll get the opposite problem, people might run into issues already fixed in the latest version.

eregon:
> And I disagree the default version is more tested, sometimes the default version is only cherry-picked by ruby-core a few days days before releases.

Right, this is different for CRuby. For TruffleRuby I can definitely say the default version is more tested (e.g. I use/"test" it a lot more than "all newer Bundler versions"). And since it is a version used in a CRuby release, potential issues are typically well known or also seen in local development.

> If you change this, you'll get the opposite problem, people might run into issues already fixed in the latest version.

[...]

PR #362, opened 2022-07-26
Use the Bundler shipped with that Ruby by default

eregon:
* Fixes https://github.com/ruby/setup-ruby/issues/358

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,9 +155,12 @@ should be able to fix them by setting `rubygems: 3.0.0` or higher.
 
 ### Bundler
 
-By default, if there is a `Gemfile.lock` file (or `$BUNDLE_GEMFILE.lock` or `gems.locked`) with a `BUNDLED WITH` section,
-that version of Bundler will be installed and used.
-Otherwise, the latest compatible Bundler version is installed (Bundler 2 on Ruby >= 2.4, Bundler 1 on Ruby < 2.4).
+By default, Bundler is installed as follows:
+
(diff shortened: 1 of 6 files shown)


eregon:
Something that may be surprising is this results in (if there is no Gemfile.lock):
* Ruby < 2.3: installs latest Bundler 1 (Bundler 2 needs Ruby 2.3+)
* Ruby 2.3 - 2.5: installs latest Bundler 2
* Ruby 2.6: uses default gem Bundler 1 (previously: installs latest Bundler 2)
* Ruby 2.7+: uses default gem Bundler 2

dentarg:
Someone might argue that this change should require a major version bump of the action. I can empathize with that seeing the fallout at https://github.com/ruby/setup-ruby/issues/358#issuecomment-1195899304, but I see your are tweaking this (https://github.com/ruby/setup-ruby/pull/363) so we might be okay in the end?

dentarg:
I understand why we want to avoid major bumps, it's a lot of churn for many where the change might not matter. There's also not nice to cut of all v1 users from future updates, that's probably the main argument.

eregon:
Yes, I thought a bit about a v2.
I think in general I want something that is compatible enough, hence a v2 is not really needed.

And if that was v2 then we'd get in a situation where:
* Few people use v2, and annoying lots of users to update for almost no reason
* Many people use v1 but I have no time to maintain more than 1 branch or backport things to it. We'd get lots of issues about v1 which is wasting time.

So a v2 seems a bad trade-off to me.
If it turns out too incompatible, we can revert this change entirely.

Issue #292, opened 2022-02-09
Allow Bundler's deployment mode setting to be overridden


nicholaspufal:
Hey there, I would like to re-surface [this discussion](https://github.com/ruby/setup-ruby/issues/153) which has been closed.

Quick recap that at the moment when using:
```
with:
  bundler-cache: true # runs 'bundle install' and caches installed gems automatically
```
Bundler will run in deployment mode "true"

I recommend re-considering a mechanism to toggle that and be able to set it to "false". 

This inability to set it to "false" is currently causing problems in a GitHub Actions workflow we have in our org. 

We have dependabot setup and in any PRs created through that automation the job that's using `ruby/setup-ruby` breaks with exit code 16 coming from Bundler. That particular job doesn't care if there is a mismatch between Gemfile and Gemfile.lock — it serves another purpose and only uses Bundler to install a needed dependency. 

Like others suggested in that other thread, currently there doesn't seem to be a way to override this behavior. I've tried to set the `BUNDLE_DEPLOYMENT` env var but `bundle config --local deployment true` defined from `ruby/setup-ruby` takes precedence.

eregon:
Deployment mode is only if there is a checked-in Gemfile.lock BTW.

Could you post a link of such a dependent PR failing?

nicholaspufal:
> Deployment mode is only if there is a checked-in Gemfile.lock BTW.
We commit the lock file in our repos to have a more deterministic environment

> BTW it should just work with dependabot if it correctly updates the Gemfile.lock.
That's the exact underlying issue here and why I think it makes sense to allow that flag to be overridden.

We have seen Dependabot making unexpected changes to Gemfile.lock [1] and that behavior + this deployment mode always set to true they create a situation where one CI job can't run `bundle install` to install a dependency (the problem compounds because that's a job required in order to merge). 

I could unblock us by not leaning on `bundler-cache: true` but like the README in this repo suggests, caching gems can be a tricky thing so would rather not go that route if possible.

While I understand the root cause here is Dependabot there should be an easy way to unblock users of this GH action even though "deployment mode" is a sensible default. Users who opt to do that would be doing so consciously through an env variable that sets it to false for instance.


[1] To expand a bit on this issue Dependabot is causing these breaking changes to `Gemfile.lock`. The working lock file looks like this:
```
GEM
  remote: https://some.repository/gems/
  remote: https://another.repository/gems/
  specs:
    # list of gems
```

[...]

PR #365, opened 2022-08-03
Add option to disable deployment mode

Pandapip1:
Pretty self-explanatory. My use-case is that I am developing on a windows machine but that the GitHub action is running on a Linux machine.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -26,7 +26,7 @@ inputs:
     default: 'false'
   deployment:
     description: 'Use deployment mode when running "bundle install". Either true or false'
-    default: 'default'
+    default: 'true'
   working-directory:
     description: 'The working directory to use for resolving paths for .ruby-version, .tool-versions and Gemfile.lock.'
   cache-version:
--- a/action.yml
+++ b/action.yml
@@ -24,6 +24,9 @@ inputs:
   bundler-cache:
     description: 'Run "bundle install", and cache the result automatically. Either true or false.'
     default: 'false'
+  deployment:
+    description: 'Use deployment mode when running "bundle install". Either true or false'
+    default: 'default'
   working-directory:
     description: 'The working directory to use for resolving paths for .ruby-version, .tool-versions and Gemfile.lock.'
   cache-version:
--- a/bundler.js
+++ b/bundler.js
@@ -142,8 +142,8 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
 
   await exec.exec('bundle', ['config', '--local', 'path', bundleCachePath], envOptions)
 
-  if (fs.existsSync(lockFile) && core.getInput('deployment') === 'true') {
-    await exec.exec('bundle', ['config', '--local', 'deployment', 'true'], envOptions)
+  if (fs.existsSync(lockFile)) {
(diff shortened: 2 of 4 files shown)


dentarg:
Linking with past discussion about deployment mode: https://github.com/ruby/setup-ruby/issues/292 (see https://github.com/ruby/setup-ruby/issues/292#issuecomment-1050924495 for a workaround)

Also sounds like you should be able to use deployment mode if you add Linux as a platform? (And there's work in progress for Bundler making this easier (despite the title of the now draft-PR) https://github.com/rubygems/rubygems/pull/5700)

Pandapip1:
The thing is, the gemfile *isn't* out of sync and installs fine with no changes.

dentarg:
Sounds like a Bundler bug? Can you share the specifics so other can reproduce?

Pandapip1:
I can point you to a commit: https://github.com/ethereum/EIPs/tree/17b8bf36950848eca4482f3488b4ccc5ea130b8a

MSP-Greg:
There are a lot of Bundler config options.  Obviously, adding even a small number of them to this action's inputs would be a PITA.

Most can be set via ENV variables.

At present, setup-ruby is setting the following:
```
bundle config --local deployment true
```

Maybe a better solution is to set it to the equivalent of `ENV.fetch('BUNDLE_DEPLOYMENT', 'true')`?

Update the README to discuss setting Bundler options via ENV, and mention the default value for BUNDLE_DEPLOYMENT?

As to the PR from ethereum/EIPs, see https://github.com/ethereum/EIPs/pull/5383

@Pandapip1 I would be very careful pinning Actions to a particular commit, as in:
```
actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
ruby/setup-ruby@08245253a76fa4d1e459b7809579c62bd9eb718a
```

Pandapip1:
> Maybe a better solution is to set it to the equivalent of ENV.fetch('BUNDLE_DEPLOYMENT', 'true')?

I would be happy with that.

> I would be very careful pinning Actions to a particular commit, as in:

The reason we are doing so is to protect against supply chain attacks.

MSP-Greg:
> to protect against supply chain attacks

And you're using `ruby-version: 2.6.0`?

eregon:
https://github.com/ethereum/EIPs/runs/7599734447?check_suite_focus=true
```
 You are trying to install in deployment mode after changing
  your Gemfile. Run `bundle install` elsewhere and add the
  updated Gemfile.lock to version control.
  
  If this is a development machine, remove the /home/runner/work/EIPs/EIPs/Gemfile
  freeze
  by running `bundle config unset deployment`.
  
  You have deleted from the Gemfile:
  * wdm (~> 0.1.0)
```
So clearly they are out of sync (or it could be a Bundler bug maybe).
Either way, that tell me the default is good, and make people investigate what they do wrong with the Gemfile/Gemfile.lock.
Disabling this would more or less be the same as not having a lockfile.

eregon:
Yes, looking if `BUNDLE_DEPLOYMENT=false` could be one way, but so far I have yet to see a valid use case for it, in my experience it reveals important issues with Gemfile.lock and there is no need to disable it.

PR #369, opened 2022-08-10
README.md updates & 'use ruby/setup-ruby@v1

MSP-Greg:
I thought a note about the 'available Ruby version' issue would be helpful at the top of the document.

It does overlap with the [VERSIONING paragraph](https://github.com/ruby/setup-ruby/blob/699560a00d6008d3844c932e3e15c404aaad9508/README.md#versioning)...


eregon:
People should already be aware of this via the issue template (https://raw.githubusercontent.com/ruby/setup-ruby/master/.github/ISSUE_TEMPLATE/bug_report.md).
I think this will not help more, people who don't know how to read an issue template probably can't read READMEs either.
For such people, I think it is far to close with just a link to https://github.com/ruby/setup-ruby#versioning and nothing else.

Homebrew goes further and says it bans people not respecting the template, I think that's maybe too much.

README.md
```diff
@@ -37,15 +45,15 @@ which means Ruby ≤ 2.4 is unmaintained and considered insecure.
...
-| Operating System | Recommended | Other Supported Versions |
-| ----------- | -------- | -------- |
-| Ubuntu  | `ubuntu-latest`  (= `ubuntu-20.04`) | `ubuntu-18.04`, `ubuntu-22.04` |
-| macOS   | `macos-latest`   (= `macos-11`)  | `macos-10.15`, `macos-12` (beta) |
-| Windows | `windows-latest` (= `windows-2022`) | `windows-2019` |
+| Operating System | Recommended                         | Other Supported Versions         |
+| ---------------- | ----------------------------------- | -------------------------------- |
+| Ubuntu           | `ubuntu-latest`  (= `ubuntu-20.04`) | `ubuntu-18.04`, `ubuntu-22.04`   |
+| macOS            | `macos-latest`   (= `macos-11`)     | `macos-10.15`, `macos-12` (beta) |
+| Windows          | `windows-latest` (= `windows-2022`) | `windows-2019`                   |
```
eregon, line 52: I don't want to manually maintain alignment here, it means lots of extra lines to change.

eregon:
Maybe we should use an Issue Form template like https://github.com/Homebrew/brew/blob/master/.github/ISSUE_TEMPLATE/bug.yml

README.md
MSP-Greg: So don't.  It's a lot more readable than the current.

MSP-Greg:
Having the info right at the top so one doesn't even need to scroll both here and in the [Marketplace](https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby) will possibly mean people don't need to see the issue template...

eregon:
> Having the info right at the top so one doesn't even need to scroll both here and in the [Marketplace](https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby) will possibly mean people don't need to see the issue template...

My goal is certainly not to save a few seconds to people who can't understand the trade-off of pinning to a sha and think about it.
I think the main issue is the text would link directly to filing an issue, that's been removed, then people will likely give it a thought before filing an issue for a new version.

eregon (approved):
I pushed to reduce the changes and reword a little bit.

PR #372, opened 2022-08-17
Add JRuby v9.3.7.0

fallwith:
Add v9.3.7.0 to the list of supported JRuby versions


eregon:
I'll use https://github.com/ruby/setup-ruby/pull/373 instead because there is an extra newline in `dist/index.js` which would fail CI.

eregon:
https://github.com/ruby/setup-ruby/releases/tag/v1.115.0

PR #375, opened 2022-08-22
Set default value in action.yml.

M-Yamashita01:
Until now, the default value was not specified in action.yml, only in the index.js.
Therefore, it was hard to find what was in the default value.
To solve this problem, we add the default value in action.yml.
And we delete the default value in index.js because the default value is specified in action.yml.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -14,6 +14,7 @@ inputs:
       For 'default', no action is taken and the version of RubyGems that comes with Ruby by default is used.
       For 'latest', `gem update --system` is run to update to the latest RubyGems version.
       Similarly, if a version number is given, `gem update --system <version>` is run to update to that version of RubyGems, as long as that version is newer than the one provided by default.
+    default: 'default'
   bundler:
     description: |
       The version of Bundler to install. Either 'Gemfile.lock' (the default), 'default', 'latest', 'none', or a version number (e.g., 1, 2, 2.1, 2.1.4).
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -9,14 +9,14 @@ const bundler = require('./bundler')
...
 const windows = common.windows
 
-const inputDefaults = {
-  'ruby-version': 'default',
-  'rubygems': 'default',
-  'bundler': 'Gemfile.lock',
-  'bundler-cache': 'false',
-  'working-directory': '.',
-  'cache-version': bundler.DEFAULT_CACHE_VERSION,
-}
```
eregon, line 19: We cannot remove this, this is useful for actions like https://github.com/ruby/setup-ruby-pkgs.

action.yml
```diff
@@ -14,22 +14,28 @@ inputs:
...
     default: 'false'
   working-directory:
     description: 'The working directory to use for resolving paths for .ruby-version, .tool-versions and Gemfile.lock.'
+    default: '.'
   cache-version:
     description: |
       Arbitrary string that will be added to the cache key of the bundler cache. Set or change it if you need
       to invalidate the cache.
+      For '0', enable the bundler cache.
+      For any value other than '0', ignore contents of the cache and get and build all the gems anew.
```
eregon, line 37: This is incorrect, can you remove?

eregon:
There is a disadvantage to specify defaults like this, and this is a big reason why they not set, on purpose:
Compare this run on master:
https://github.com/ruby/setup-ruby/runs/7938241309?check_suite_focus=true
```
Run ./
  with:
    ruby-version: 1.9
    bundler-cache: true
```
with
https://github.com/ruby/setup-ruby/runs/7949177099?check_suite_focus=true
```
Run ./
  with:
    ruby-version: 2.1
    bundler-cache: true
    rubygems: default
    bundler: Gemfile.lock
    working-directory: .
    cache-version: 0
```

IMHO the former looks a lot cleaner, and represents more accurately what inputs were actually provided by the user/what it looks like in workflow YAML.

> Therefore, it was hard to find what was in the default value.

They are pretty clearly mentioned in the description for `rubygems` & `bundler`, so I think it's not hard.
For working-directory/cache-version they are not mentioned in the description but it seems unnecessary.

PR #376, opened 2022-08-22
Don't install Bundler on Windows mingw, mswin & ucrt builds

MSP-Greg:


MSP-Greg:
Something is odd with the Windows 2019 20220814.1 image, as it fails.

But, I had a CI run checking master a few hours earlier using 20220821.1, and it [passed](https://github.com/MSP-Greg/setup-ruby/runs/7956183351?check_suite_focus=true#step:1:9).

eregon:
I changed the check a little bit, should be equivalent.

Those failures are very strange: https://github.com/ruby/setup-ruby/runs/7960716852?check_suite_focus=true
I'll wait CI.

eregon:
Ah, it's because it's already failing on master, `which rake` doesn't work :/
I'll adapt the test I guess.

eregon:
All green now :)

eregon (approved):

PR #386, opened 2022-10-13
Update dependencies using yarn upgrade

### LINKED ISSUE #385: Deprecation warning for set-output command in GitHub Actions
Hello team,

It appears that GitHub is planning to disable `set-output` command soon according to the [blog](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

I am currently using the follow code to setup Ruby in my GitHub Actions:
```yaml
- name: Setup Ruby  
   uses: ruby/setup-ruby@v1
```

Warning message:

`Warning: The set-output command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/`

smortex:
GitHub deprecated some commands that are used by older versions of the
GitHub action dependencies.  This result in deprecation warnings when
the task is run and was reported as #385.

Fixes #385

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -664,6 +664,9 @@ function saveCache(paths, key, options) {
         const cachePaths = yield utils.resolvePaths(paths);
         core.debug('Cache Paths:');
         core.debug(`${JSON.stringify(cachePaths)}`);
+        if (cachePaths.length === 0) {
+            throw new Error(`Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.`);
+        }
         const archiveFolder = yield utils.createTempDirectory();
         const archivePath = path.join(archiveFolder, utils.getCacheFileName(compressionMethod));
         core.debug(`Archive Path: ${archivePath}`);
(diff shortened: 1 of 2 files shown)


nobu:
I tried `yarn upgrade && yarn run package`.
yarn.lock is same, but dist/index.js is smaller about 2300 lines than yours.
From where did the lines from `/***/ 3532:` before `/***/ 7371:` come?

smortex:
Hum, it's basically what I did on my $WORK machine: I do not do JS development so I used the `node:latest` docker image to run these commands (from memory I run something like `docker run -it --mount src=$PWD,target=/setup-ruby,type=bind --entrypoint /bin/bash node`) :shrug: .

eregon:
Done in https://github.com/ruby/setup-ruby/commit/eae47962baca661befdfd24e4d6c34ade04858f7.
I also had a slightly different diff like nobu said.

PR #388, opened 2022-10-14
Automatically update yarn packages

nobu:


.github/workflows/update.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Check diffs
+        id: diff
+        run: git diff --no-ext-diff --ignore-submodules --stat --exit-code
+        continue-on-error: true
+      - name: Commit
+        run: |
+          git pull --ff-only origin ${GITHUB_REF#refs/heads/}
+          git add --all
+          git commit --message="Update packages at ${TODAY}"
+          git push origin ${GITHUB_REF#refs/heads/}
```
eregon, line 42: I don't think I would want something that pushes automatically to master, but rather it creates a PR or something.

Also it means the permissions cannot just be:
```
permissions:
  contents: read
```
and so it's a lot harder to assess how safe is this or not (it seems pretty unsafe at first glance, any dep update could read the GitHub token etc).

eregon:
I think for now I prefer to explicitly update dependencies, only when needed.
That seems safer to me from a security perspective.

PR #399, opened 2022-11-14
Update Packages (tool-cache, http-client) & regenerate dist/index.js

MSP-Greg:


MSP-Greg:
@eregon

Something is odd.  Windows-2019 Ruby 2.0.0 shows rake 13.0.6 (see [log](https://github.com/ruby/setup-ruby/actions/runs/3457650272/jobs/5771290010#step:3:49)), but all the other Ruby 2.0.0 jobs show `Using rake 12.3.3`?

eregon (approved):

MSP-Greg:
That could be, but only on Windows 2.0.0?  Regardless, the Windows job passed when merged, but a 2.1 job failed.

Does the readme say anything equivalent to "Don't even think about opening an issue about old Rubies that no one should be using"...

Sorry,  rant off.

Issue #364, opened 2022-07-28
[Idea] Use latest stable ruby if no `with - ruby-version` specified


ShockwaveNN:
I got a little idea - I got a little CI task that check that `yardoc` are generating correct and full documentation for my gem

I don't need any specific version of ruby in this task, so I fought if I init CI task with this config:

```
name: Documentation
on: [push]
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Set up latest ruby
        uses: ruby/setup-ruby@v1
      - name: Install gem dependencies
        run: |
          gem install bundler
          bundle install --jobs 4 --retry 3
      - name: Check that yardoc is correct
        run: yardoc . --fail-on-warning > yardoc.log
      - name: Check that code 100% documented
        run: cat yardoc.log | grep -q '100.00% documented'
 ```

It will be using the latest stable ruby (since I don't specify any)

But instead, this task failed with
```
Run ruby/setup-ruby@v1
Error: Error: input ruby-version needs to be specified if no .ruby-version or .tool-versions file exists
```

This error is 100% clear for me, but maybe this behavior should be reconsidered and in that case -just latest stable ruby should be used, like v3.1 right now

eregon:
Use `ruby-version: ruby` if that's what you want (the latest CRuby release).

eregon:
In general that's a little bit more risky though because e.g. yardoc might not support the latest CRuby release, especially when a new minor comes out. So I don't think we want this as a sort of "default".

ShockwaveNN:
@eregon I agree with you latest point, but I'm ready for this risk and don't want to update this task with each new release of ruby (even once a year)

> 
Use ruby-version: ruby if that's what you want (the latest CRuby release).

I'll check that and if so - It work for me, but I didn't find any mention of this in documentation (prove me wrong if I'm a little bit blind ))

eregon:
It's documented in https://github.com/ruby/setup-ruby#supported-version-syntax (engine-only)

ShockwaveNN:
Oops, sorry, agree with you

PR #417, opened 2022-12-13
Update README.md

kojix2:
Hi setup-ruby maintainers!

> Added example of using the latest stable version of ruby
> Related to #364

When creating documentation or using other Ruby tools, you may want to specify the latest stable version of CRuby. In such cases, you can specify `ruby-version` with `ruby`, but it can be a little difficult to find this information from the README. Therefore, I thought it would be helpful to provide an example of how to specify ruby in the README, to make it easier to find the method you are looking for. I myself found this solution by searching for "latest stable ruby" in the issues, not from the README. Thank you for always maintaining setup-ruby. Best regards.


eregon (approved):
Thanks!

PR #418, opened 2022-12-19
ubuntu-latest is now 22.04

fa11enangel:
According to GitHub Blog ubuntu-latest is now 22.04.

see: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/


eregon (approved):
Thanks. Probably I should change this to not mention `*-latest` because that's just aliases anyway.

Issue #422, opened 2022-12-24
Ruby 2.3 to 2.5 can't install bundler


MaxLap:
### Ensure the following before filing this issue

- [X] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [X] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

### Are you running on a GitHub-hosted runner or a self-hosted runner?

GitHub-hosted runner

### The workflow code or a link to the workflow file

```yaml
https://github.com/MaxLap/activerecord_where_assoc/blob/master/.github/workflows/run_tests.yml
```

### Link to the log of a failed workflow job, or to a gist with the output

https://github.com/MaxLap/activerecord_where_assoc/actions/runs/3771542199/jobs/6412280221

### The command and output of the failing step

```shell
On a Ruby 2.3 to 2.5 installation (I tested the command with ruby 2.4.10p364 locally and have the same problem):


gem install bundler -v '~> 2.0'
ERROR:  Error installing bundler:
	The last version of bundler (~> 2.0) to support your Ruby & RubyGems was 2.3.26. Try installing it with `gem install bundler -v 2.3.26`
	bundler requires Ruby version >= 2.6.0. The current ruby version is 2.4.10.364.
```
```


### Any other notes?

I know since I can reproduce locally, you say it's not a setup-ruby bug, but I think in this case, it is, since it seems setup-ruby is using the wrong command in that case.

Bundler 2.4.0 was just released, which has a minimum Ruby version of 2.6.

The default bundle installation command of `gem install bundler -v '~> 2.0'` doesn't work on those Ruby version because it seems `gem install` only uses the `~= 2.0` to pick the latest version, not necessarily one that is compatible with the current Ruby.

deivid-rodriguez:
Bundler 2.4 has dropped support for Ruby 2.3, Ruby 2.4 and Ruby 2.5. But the RubyGems version provided with these rubies cannot handle `gem install bundler -v '~> 2.0'` well, if the latest version of Bundler does not support these rubies. Like it started happening now.

The solution is that `ruby/setup-ruby` provides at least RubyGems 3.2.3 by default on these rubies.

[...]

PR #423, opened 2022-12-24
Bundler-2.4 requires Ruby 2.6 or later

hsbt:
Resolved https://github.com/ruby/setup-ruby/issues/422

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -107,6 +107,7 @@ export async function installBundler(bundlerVersionInput, rubygemsInputSet, lock
   }
 
   // Use Bundler 1 when we know Bundler 2 does not work
+  // Use Bundler 2.3 when we use Ruby 2.3.2-2.5
   if (bundlerVersion.startsWith('2')) {
     if (engine === 'ruby' && floatVersion <= 2.2) {
       console.log('Bundler 2 requires Ruby 2.3+, using Bundler 1 on Ruby <= 2.2')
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Thanks for a PR on Christmas Eve (Happy Holidays/Merry Christmas!)

Not my call, but...

 I think it would be better to force install of RG 3.2.3 (which correctly handles `required_ruby_version`), rather than force a bundler version?  I think that can be done such that it leaves RUby 2.6 and later alone.  I think I can do that quickly...

MSP-Greg:
Thanks for a PR on Christmas Eve (Happy Holidays/Merry Christmas!)

Not my call, but...

 I think it would be better to force install of RG 3.2.3 (which correctly handles `required_ruby_version`), rather than force a bundler version?  I think that can be done such that it leaves RUby 2.6 and later alone.  I think I can do that quickly...

eregon (approved):
Thank you for the PR!

hsbt:
Hi Greg, Eregon, Happy Holidays and Thanks for all your works for Ruby.

>Would it be better to force install of RG 3.2.3 (which correctly handles required_ruby_version), rather than force a bundler version?

It's not needed yet. Because RubyGems is internally usage different with Bundler by Ruby users.

eregon:
@MSP-Greg That would choose the same Bundler version, but also update RubyGems which users unlikely have done locally to the same version + a more risky change

eregon:
https://github.com/ruby/setup-ruby/releases/tag/v1.129.0

PR #427, opened 2022-12-25
Bundler higher than ~> 2.4.0 will require Ruby 3.0 or later

deivid-rodriguez:
This is so that I don't forget to do this before next year's Bundler release.

I also fixed a small bug introduced with #423 where an explicit and valid input configured by users like `bundler: 2.2` would be ignored and "bundler: 2.3" would be installed instead.


eregon (approved):
Thank you! This looks great.

hsbt:
👍 Thanks!

PR #429, opened 2022-12-26
Reenable fixed RubyGems test

deivid-rodriguez:
Use the same Ruby as the previous test, since it's very related, but use a newer RubyGems to force the upgrade case.


eregon (approved):
Thanks!

PR #431, opened 2022-12-27
Windows - updates for using OpenSSL 3 in Ruby 3.2 & head

MSP-Greg:
The RubyInstaller2 builds of 3.2 and 'head' are using a custom OpenSSL 3 package.

[ruby/setup-msys2-gcc](https://github.com/ruby/setup-msys2-gcc), which assembles archive files of Windows build tools, has been updated and now creates archives (named with suffixes) that allow for different OpenSSL versions.  See the [release list](https://github.com/ruby/setup-msys2-gcc/releases/tag/msys2-gcc-pkgs).

This PR adds code to use the correct archive file.

See ['Windows, OpenSSL, Ruby 3.2' discussion](https://github.com/ruby/setup-ruby/discussions/428).

This code will require changes when MSYS2 upgrades from using OpenSSL 1.1.1 to 3.0.x.

Note that the `mswin` archive is only used with the 'mswin' head build, and it already uses OpenSSL 3.0.x.

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -1,6 +1,8 @@
 // 7z arguments
 //   -aoa overwrite existing, -bd disable progress indicator
 
+// OpenSSL version detection is needed in installGCCTools and installJRubyTools
+
 const fs = require('fs')
 const path = require('path')
 const cp = require('child_process')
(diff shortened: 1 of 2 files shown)


windows.js
```diff
@@ -129,7 +140,8 @@ async function installMSYS2Tools() {
 // windows-2022 and later images
 export async function installJRubyTools() {
   await installMSYS2Tools()
-  await installGCCTools('mingw64')
+  // Will need to add OpenSSL version detection when JRuby uses OpenSSL 3.0 ?
```
eregon, line 143: I think JRuby never uses libssl, but rather JDK stuff or bouncycastle (I think both not using libssl)

eregon (approved):
Thanks!

PR #435, opened 2022-12-29
Add RubyInstaller-3.2.0-1 for Windows

larskanis:
This might be combined with https://github.com/ruby/setup-ruby/pull/431 .


larskanis:
Here is the release post: https://rubyinstaller.org/2022/12/29/rubyinstaller-3.2.0-1-released.html

MSP-Greg:
JFYI, OpenSSL 3.0.7 is available in [setup-msys2-gcc](https://github.com/ruby/setup-msys2-gcc/actions/runs/3803106374/jobs/6469228333#step:4:110).

As soon as this and #431 are added, Puma will be running CI on Ruby 3.2.0, and it compiles with OpenSSL.  Note that it's already using OpenSSL 3.0.7 with the Ruby mswin build, as that's the version from MSFT/vcpkg...

PR #460, opened 2023-02-07
Remove debug information again

lobner:
Remove debug information from 8aa012b56d9eee979b6da5b9f679037c9a5e9a4d


eregon:
This is useful, so I keep it. It's an internal script anyway.

PR #467, opened 2023-02-22
Removed macos-10.15 because it will be removed at 2023-03-31

hsbt:
I got the notification from GitHub because we still have configuration of `macos-10.15` at GitHub Actions. We should remove it from `test.yml` at least.

But I'm not sure we also removed it from `setup-ruby` at this time.


eregon:
Similarly Ubuntu 18.04 is being deprecated and has brownout periods, probably we should remove both since it's no longer possible to reliably build on those platforms with GH runners.
It seems best to remove both at the same time/for the same setup-ruby release.

Could you update this PR to also remove Ubuntu 18.04?

hsbt:
Sure, I will also remove `ubuntu-18.04`.

eregon:
Links for the record:
https://github.com/actions/runner-images/issues/5583
https://github.com/actions/runner-images/issues/6002

eregon (approved):
Thank you!

hsbt:
Thanks for merging 🚀

PR #470, opened 2023-03-01
Update example to remove unsupported versions

AriehSchneier:


eregon:
Thanks, this code & docs just got removed in https://github.com/ruby/setup-ruby/pull/473 so there is no need to change anything now.

Issue #242, opened 2021-12-10
Tool cache path on self-hosted runners can't be hardcoded


Vyazovoy:
Hi! I use self-hosted runners and I have several instances on the same physical machine. I see that this action uses hardcoded path to tool cache but there is no guarantee that even when people have only one self-hosted runner that it will be `/Users/runner...`. Would it be possible to avoid all path hardcoding?

MSP-Greg:
I believe the code used is below.  The hardcoded values match `ENV['RUNNER_TOOL_CACHE']`.  Would that work in your situation?

See most recent 'Image Info' run in https://github.com/MSP-Greg/actions-image-testing/actions.  Currently is https://github.com/MSP-Greg/actions-image-testing/actions/runs/1492346296, see 'Show ENV' step.

https://github.com/ruby/setup-ruby/blob/471466b54cb9744f371b61fb5ca33975e116fefb/common.js#L126-L137

MSP-Greg:
@Vyazovoy

This is problematic with self-hosted runners.  Two cases:

1. You're using a Ruby that is contained in your runner's toolcache.  I believe using `ENV['RUNNER_TOOL_CACHE']` would work.

2. You want to download and install one of the pre-built Rubies (ie, not one in your toolcache).  I believe the path is hardcoded for all Rubies except the Windows builds.  Not sure about Window non-MRI builds.

So, I suppose the code could check for a match in the toolcache at `ENV['RUNNER_TOOL_CACHE']`, and if a match is not found, install to the hardcoded path...

Vyazovoy:
Thanks for the explanation, @MSP-Greg 🙏 [Here](https://github.com/ruby/setup-ruby/blob/master/ruby-builder.js#L22-L23) the code checks for a match in the toolcache so I should probably populate the cache manually to make this action working but it would be good to have a configurable parameter that would allow to force installation with `ruby-build` instead of downloading the prebuilt Rubies. By using that parameter and non-hardcoded toolcache path the action would install a Ruby version with `ruby-build` when called first time on a self-hosted runner and all subsequent runs would use the cached version. Also, if it is possible to detect that a runner is self-hosted, that would be good to apply the behavior automatically.

MSP-Greg:
> several instances on the same physical machine.

That certainly seems possible.  Question - do they run concurrently/at the same time?  If not, could the default toolcache path be a symlink?

Not.enough.coffee.yet...

Vyazovoy:
[...]

PR #474, opened 2023-03-03
Consider a different $RUNNER_TOOL_CACHE than the default as a self-hosted runner and add self-hosted input

eregon:
Fixes https://github.com/ruby/setup-ruby/issues/242

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -30,6 +30,11 @@ inputs:
     description: |
       Arbitrary string that will be added to the cache key of the bundler cache. Set or change it if you need
       to invalidate the cache.
+  self-hosted:
+    description: |
+      Consider the runner as a self-hosted runner, which means not using prebuilt Ruby binaries which only work
+      on GitHub-hosted runners or self-hosted runners with a very similar image to the ones used by GitHub runners.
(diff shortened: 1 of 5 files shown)


eregon:
On the deprecated and no longer supported ubuntu-18.04 it looks like:
https://github.com/eregon/setup-ruby-test/actions/runs/4325551205/jobs/7551857499
```
Error: The current runner (ubuntu-18.04-x64, RUNNER_TOOL_CACHE=/opt/hostedtoolcache) was detected as self-hosted because the platform does not match a GitHub-hosted runner image (or that image is deprecated and no longer supported).
In such a case, you should install Ruby in the $RUNNER_TOOL_CACHE yourself, for example using https://github.com/rbenv/ruby-build
You can take inspiration from this workflow for more details: https://github.com/ruby/ruby-builder/blob/master/.github/workflows/build.yml
$ ruby-build 3.2.1 /opt/hostedtoolcache/Ruby/3.2.1/x64
Once that completes successfully, mark it as complete with:
$ touch /opt/hostedtoolcache/Ruby/3.2.1/x64.complete
It is your responsibility to ensure installing Ruby like that is not done in parallel.
```

PR #482, opened 2023-03-15
Update yarn.lock, remove Windows mswin ssl cert related code

MSP-Greg:
1. MSYS2 recently changed the folder that the ssl cert file is located in.  'Git for Windows' is built with MSYS2.  The most recent version has the files in `etc/ssl`, older versions have them in `ssl`.  Code now looks in both locations.  This file is needed for Windows Rubies 2.3 and earlier, as they don't include a cert file, and the code uses the cert file included with 'Git for Windows'...

2. The ruby-loco mswin build now has a bundled cert file.  Removed the code for that.

3. Updated `yarn.lock`.

Note that the Actions images that are currently rolling out will not work without this PR.  Sorry, I forgot about the 'Git for Windows' change due to the MSYS2 change...

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -720,6 +720,29 @@ function setupPath(newPathEntries) {
 
 "use strict";
 
+var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
+    if (k2 === undefined) k2 = k;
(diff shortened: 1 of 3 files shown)


eregon (approved):
Thanks!

Issue #483, opened 2023-03-17
Restoring bundler cache is not working in v1.144.1


jacob-carlborg-apoex:
### Ensure the following before filing this issue

- [X] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [X] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

### Are you running on a GitHub-hosted runner or a self-hosted runner?

GitHub-hosted runner

### The workflow code or a link to the workflow file

```yaml
name: CI
on: [push]
concurrency:
  group: ${{ github.ref }}
  cancel-in-progress: true

jobs:
  bundle:
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v3

      - name: Install dependencies
        run: |
          sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends \
            libxmlsec1-dev libxmlsec1-openssl wkhtmltopdf

      - name: Github setup
        run: |
          git config --global url."https://${{secrets.GH_ACCESS_TOKEN}}:x-oauth-basic@github.com/".insteadOf "git@github.com:"

      - uses: ruby/setup-ruby@v1.144.1
        with:
          bundler-cache: true
```

### Link to the log of a failed workflow job, or to a gist with the output

https://gist.github.com/jacob-carlborg-apoex/370a6197d4f1f5102f0a8c5e44235179

### The command and output of the failing step

```shell
> bundle install
/opt/hostedtoolcache/Ruby/2.7.2/x64/bin/bundle config --local path /home/runner/work/foobar/foobar/vendor/bundle
/opt/hostedtoolcache/Ruby/2.7.2/x64/bin/bundle config --local deployment true
Cache key: setup-ruby-bundler-cache-v5-ubuntu-20.04-x64-ruby-2.7.2-wd-/home/runner/work/foobar/foobar-with--without--Gemfile.lock-74395cbea7219758bf34c731625de123b8704e40d1e7f64d3d455f933b051a3f
/opt/hostedtoolcache/Ruby/2.7.2/x64/bin/bundle install --jobs 4
Fetching gem metadata from https://rubygems.org/.......
```
```


### Any other notes?

[...]

PR #484, opened 2023-03-17
Lock @actions/cache to 3.0.6, later versions break bundler caching

MSP-Greg:
See #483.  I can't look at it until (hopefully) this weekend.

It may be a bug with @actions/cache, as I tried both 3.1.4 & 3.2.1.  Not.sure.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,10 +2,10 @@
 # yarn lockfile v1
 
 
-"@actions/cache@^3":
-  version "3.2.1"
-  resolved "https://registry.yarnpkg.com/@actions/cache/-/cache-3.2.1.tgz#bbd4680aeab811f3e839e58be493ffe7583d0311"
-  integrity sha512-QurbMiY//02+0kN1adJkMHN44RcZ5kAXfhSnKUZmtSmhMTNqLitGArG1xOkt93NNyByTlLGAc5wIOF/dZ2ENOQ==
+"@actions/cache@3.0.6":
+  version "3.0.6"
+  resolved "https://registry.yarnpkg.com/@actions/cache/-/cache-3.0.6.tgz#b76c51a78c927fcf0eb631b96a07e34f575c422b"
+  integrity sha512-Tttit+nqmxgb2M5Ufj5p8Lwd+fx329HOTLzxMrY4aaaZqBzqetgWlEfszMyiXfX4cJML+bzLJbyD9rNYt8TJ8g==
   dependencies:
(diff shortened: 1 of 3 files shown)


MSP-Greg:
Two jobs have 429 errors?  That's 'Too Many Requests'?  What else is going to happen today???

See:
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
https://www.rfc-editor.org/rfc/rfc6585.html#section-4

MSP-Greg:
I did some more testing, and it seems that 3.1.0 and 3.1.1 also seem to work.  3.1.2 and later do not.  Also, see:

https://github.com/actions/toolkit/issues/1377

simi:
Thanks for addressing this @MSP-Greg. Would it be enough to restart failing checks for now, :pray: merge and release?

eregon:
My thoughts is we should avoid "blindly" updating to latest versions of dependencies without a need, things tend to break, it's not the first time. Thank you for the PR.

yarn.lock
```diff
@@ -1,428 +1,428 @@
...
-xmlbuilder@~11.0.0:
-  version "11.0.1"
-  resolved "https://registry.yarnpkg.com/xmlbuilder/-/xmlbuilder-11.0.1.tgz#be9bae1c8a046e76b31127726347d0ad7002beb3"
-  integrity sha512-fDlsI/kFEx7gLvbecc0/ohLG50fugQp8ryHzMTuW9vSa1GJ0XYWKnhsUx7oie3G98+r56aTQIUB4kht42R3JvA==
-
-yallist@^4.0.0:
-  version "4.0.0"
-  resolved "https://registry.yarnpkg.com/yallist/-/yallist-4.0.0.tgz#9bb92790d9c0effec63be73519e11a35019a3a72"
-  integrity sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A==
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
eregon, line 1: This file has CRLF that's why the diff is so big, I'll regenerate it with LF.

MSP-Greg:
> Maybe we can do that by using ruby/setup-ruby twice in in the same workflow job, and having some debug env var or so to fail if the cache isn't restored in the 2nd usage.

Yes, I thought something similar.  I back to testing this because I'm just a little frustrated by it...

PR #485, opened 2023-03-21
Fix cache with @actions/cache

MSP-Greg:
See also https://github.com/actions/toolkit/pull/1378.

If the above PR is accepted, the change to bundler.js isn't needed, but it's a good idea, as we don't want the value to change when passed to [@actions/cache's](https://github.com/actions/toolkit/tree/main/packages/cache) code...

Currently, we use:
```js
cache.restoreCache(paths, key, restoreKeys)
// then later the following may run
cache.saveCache(paths, key)
```

Current code in [@actions/cache](https://github.com/actions/toolkit/tree/main/packages/cache) modifies `paths` during the call to `cache.restoreCache`.  Since the verification hash is based on that array, `cache.saveCache` will associate a verification key with the saved cache that `cache.restoreCache` will never match...

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -236,7 +236,7 @@ async function computeBaseKey(platform, engine, version, lockFile, cacheVersion)
   const cwd = process.cwd()
   const bundleWith = process.env['BUNDLE_WITH'] || ''
   const bundleWithout = process.env['BUNDLE_WITHOUT'] || ''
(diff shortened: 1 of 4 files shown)


MSP-Greg:
Added a 2nd commit to change the `baseKey`, as I tested with essentially a flushed cache.  Or, I removed any matching caches before testing with the Puma workflow.

Please 'Squash & Merge' if you want one commit...

MSP-Greg:
Regardless, passing an object (or a non-primitive) is not a good thing to do, especially when we depend on its value later in the code...

PR #510, opened 2023-06-21
update: debian

mram-wtag:


eregon:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners doesn't have debian.

PR #542, opened 2023-10-27
Add support for macos-14

volatilegg:
## Motivation

Adding `macos-14` image to the support list to avoid error if the runner is using `macos-14`

```
Error: The current runner (macos-14-x64) was detected as self-hosted because the platform does not match a GitHub-hosted runner image (or that image is deprecated and no longer supported).
```


eregon:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources does not include macos-14.
So I guess you are using a self-hosted runner?

The problem is `macos-14` cannot be tested in CI here, maybe it works, maybe not and it cannot really be supported if it's not tested.

volatilegg:
@eregon  Thanks for the quick reply
Yes, this is blocking me on a self-hosted runner 😓 
We're trying to push the changes on a temp action, will update here once it works 
https://github.com/marketplace/actions/setup-ruby-temp?version=v.1.160.0

eregon:
I think for macOS & Windows, where in both cases we use builds on the oldest available GH-hosted runners we could let the action detect that case (self-hosted + version more recent than in GH-hosted runners) and continue, and just warn in its output about this case not being tested.

For Linux that's not an option as we have different builds for each OS version.

eregon:
BTW you don't need to publish the release to the marketplace to try your action, just a tag or branch is enough.
In fact I think it's best to avoid publishing the release to the marketplace to avoid confusion here.

volatilegg:
@eregon Oh thank you, I didn't know that :D 
And delisted the temp one on marketplace

eregon:
I'll close this because that's not the correct way to fix this.

I would welcome a PR to do https://github.com/ruby/setup-ruby/pull/542#issuecomment-1783125940 though.

icecoffin:
macOS 14 is now available on GitHub hosted runners (https://github.com/actions/runner-images/issues/7508). Could this PR be reopened and merged?

eregon:
Will be done in https://github.com/ruby/setup-ruby/pull/567 instead.

Issue #489, opened 2023-03-29
Only perform bundler actions / don't install Ruby


steveh:
We run actions with a custom Docker image that already contains Ruby and Bundler.

The [recommendation](https://github.com/actions/cache/blob/main/examples.md#ruby---bundler) is to use this action to cache gems with Bundler, but I don't want it to also install Ruby.

Can we set something like `ruby_version: none` to skip that part?

AskarZinurov:
Or maybe it will be possible to set `ruby_version: system` instead?

eregon:
I'm not clear how actions work inside Docker images, but I suppose they do?

PR welcome.
It is not completely trivial, because e.g. the Ruby version should still be part of the cached key and we need to know the Bundler major version.

AskarZinurov:
@eregon yep, seems you are right, for case in issue description, it is much clearer just run bundle install to separate bundler path and then use https://github.com/actions/cache directly. Smth like:

```
jobs:
  specs:
    runs-on: ubuntu-latest
    container: ruby:2.4.10-stretch
    services:
      redis:
        image: redis:6.2-alpine
        ports:
          - "6379:6379"
      postgres:
        image: postgres:11-bullseye
        ports:
          - "5432:5432"
    env:
      RAILS_ENV: "test"
      BUNDLE_PATH: "/opt/ruby-gemset"
    steps:
      - name: Checkout code
        uses: actions/checkout@v3
      - name: Cache gemset
        id: gemset-cache
        uses: actions/cache@v3
        with:
          path: /opt/ruby-gemset
          key: ${{ runner.os }}-${{ hashFiles('Gemfile.lock') }}
      - name: Install ruby gems
        run: bundle install --jobs 8
      - name: Create database config file
        run: cp config/database.yml.example config/database.yml
      - name: Run tests
        run: |
          bin/rake db:setup
          bin/rspec
```

steveh:
> It is much clearer just run bundle install to separate bundler path and then use https://github.com/actions/cache directly.

The linked recommendation explicitly says not to do that:

> Caching gems with Bundler correctly is not trivial and just using actions/cache is [not enough](https://github.com/ruby/setup-ruby#caching-bundle-install-manually).

> Instead, it is recommended to use ruby/setup-ruby's [bundler-cache: true option](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically) whenever possible:

eregon:
[...]

PR #548, opened 2023-12-05
Support bundler-only runs

Bo98:
#489

Seems to work across all macOS and Ubuntu runners.

Cache key has an extra `RUBY_PLATFORM` when using system Ruby as that can vary depending on how Ruby is compiled. For example, on macOS you could have a Ruby compiled for an older Darwin version than the one you are currently running on.

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -2,7 +2,9 @@ const fs = require('fs')
 const path = require('path')
 const core = require('@actions/core')
 const exec = require('@actions/exec')
+const io = require('@actions/io')
 const cache = require('@actions/cache')
+const semver = require('semver')
 const common = require('./common')
 
 export const DEFAULT_CACHE_VERSION = '0'
@@ -58,7 +60,7 @@ async function afterLockFile(lockFile, platform, engine, rubyVersion) {
   }
 }
 
-export async function installBundler(bundlerVersionInput, rubygemsInputSet, lockFile, platform, rubyPrefix, engine, rubyVersion) {
+export async function installBundler(bundlerVersionInput, rubygemsInputSet, systemRubyUsed, lockFile, platform, rubyPrefix, engine, rubyVersion) {
   let bundlerVersion = bundlerVersionInput
 
   if (rubygemsInputSet && (bundlerVersion === 'default' || bundlerVersion === 'Gemfile.lock')) {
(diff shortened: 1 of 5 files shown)


eregon:
Could you clarify what's the use-case?

It does look like a lot of changes and complications, so it would need a really good use case without workarounds to be worth it.

bundler.js
```diff
@@ -79,7 +81,21 @@ export async function installBundler(bundlerVersionInput, rubygemsInputSet, lock
   const floatVersion = common.floatVersion(rubyVersion)
 
   if (bundlerVersion === 'default') {
-    if (common.isBundler2dot2Default(engine, rubyVersion)) {
+    if (systemRubyUsed) {
```
eregon, line 84: I think we shouldn't pass `systemRubyUsed` to bundler.js, it should just work if we pass the correct `rubyVersion`.
So we'd ask the RUBY_VERSION of the system ruby early on, and use that for `rubyVersion`.
We'd just skip downloading, extracting and adding to PATH if the input is system.

Bo98:
> Could you clarify what's the use-case?
> 
> It does look like a lot of changes and complications, so it would need a really good use case (without good workarounds) to be worth it.

#489 lists some, but for me specifically I need to conditionally use a custom-compiled Ruby while keeping the standard bundler install. Workaround would be to just copy the bundler caching logic here into a personal local action for my own use.

I suppose it in part comes from `ruby/setup-ruby` being more of a `ruby/setup-ruby-and-bundler`.

If this is too much then that's fine - I'll figure something else out.

bundler.js
Bo98: Basing it on `rubyVersion` isn't quite correct as:
* Ubuntu image doesn't ship with any Bundler
* macOS 11 image ships with a default Ruby 2.7 but with Bundler 2.4.

The logic used here for system Ruby is generic enough however that we could replace the existing `common.isBundler2dot2Default` logic here and thus avoid the `systemRubyUsed`.

eregon:
> #489 lists some,

I only see Docker there and I wonder if actions even work at all within Docker images.

> but for me specifically I need to conditionally use a custom-compiled Ruby while keeping the standard bundler install.

I think one way would be to install your custom-compiled Ruby under the toolcache.
Then it would be found by this action and no changes should be needed.
See the end of https://github.com/ruby/setup-ruby/blob/master/README.md#using-self-hosted-runners, notably that self-hosted true mode.

bundler.js
Bo98: > Which Ubuntu image? IIRC `ubuntu-22.04` ships with some Ruby version in PATH by default from the toolcache, and that has Bundler.

This is what I get with `ubuntu-22.04`: https://github.com/Bo98/workflow-test/actions/runs/7118996527/job/19383150052

macOS and Ubuntu runners do not use a toolcache Ruby by default. They use ones from system package managers.

Bo98:
> I think one way would be to install your custom-compiled Ruby under the toolcache.

I can't move the install but maybe could maybe nuke the toolcache and symlink? Seems very hacky but I haven't tried it yet.

invisiblepancake (commented):
Files Modified:

    .github/workflows/test.yml
        Added system Ruby version to the matrix.
        Added steps to output Ruby prefix and conditional checks for system Ruby version.

    action.yml
        Updated the description for ruby-version input to include system option.

    bundler.js
        Introduced logic to handle system Ruby version.
        Added checks and handling for system Bundler version.
        Updated caching mechanism to include system Ruby platform.

    dist/index.js
        Updated to support system Ruby version.
        Added handling for system Bundler and Ruby prefix.

    index.js
        Added support for system Ruby version.
        Updated logic to handle system Bundler and Ruby prefix.

Overall, the commit introduces support for running with the system Ruby version and includes adjustments to workflows and scripts to handle this new option.

invisiblepancake (commented):
Files Modified:

    .github/workflows/test.yml
        Added system Ruby version to the matrix.
        Added steps to output Ruby prefix and conditional checks for system Ruby version.

    action.yml
        Updated the description for ruby-version input to include system option.

    bundler.js
        Introduced logic to handle system Ruby version.
        Added checks and handling for system Bundler version.
        Updated caching mechanism to include system Ruby platform.

    dist/index.js
        Updated to support system Ruby version.
        Added handling for system Bundler and Ruby prefix.

    index.js
        Added support for system Ruby version.
        Updated logic to handle system Bundler and Ruby prefix.

Overall, the commit introduces support for running with the system Ruby version and includes adjustments to workflows and scripts to handle this new option.

eregon:
> I can't move the install but maybe could maybe nuke the toolcache and symlink? Seems very hacky but I can give it a try.

Yes, I think adding your custom Ruby to the toolcache is the best way here.

Also this PR is really big and seems hard to maintain for this niche feature so it seems best to not merge it.

Issue #529, opened 2023-10-02
Runner cannot update rubygems


luke-hill:
### Ensure the following before filing this issue

- [X] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [X] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

### Are you running on a GitHub-hosted runner or a self-hosted runner?

GitHub-hosted runner

### Link to the failed workflow job (must be a public workflow job, so the necessary information is available)

https://github.com/cucumber/cucumber-ruby-core/actions/runs/6383187734/job/17323334688?pr=265

PR code quicklink: https://github.com/cucumber/cucumber-ruby-core/pull/265/files

### Any other notes?

Possibly related to previous issue I raised #518 - The fix I was told was to update rubygems. I now get an error when I try update on ruby 2.5

```
ERROR:  Error installing rubygems-update:
  	There are no versions of rubygems-update (= 3.4.20) compatible with your Ruby & RubyGems
  	rubygems-update requires Ruby version >= 2.6.0. The current ruby version is 2.5.0.
  ERROR:  While executing gem ... (NoMethodError)
      undefined method `version' for nil:NilClass
```

eregon:
That's clearly a bug of RubyGems or that old Ruby 2.5.
Probably `gem update --system` doesn't work on Ruby 2.5, did you try that locally?
You can probably workaround by manually passing the last RubyGems version which supports Ruby 2.5.
See https://github.com/ruby/setup-ruby/issues/441

In general I would recommend to avoid updating rubygems or bundler and just use the defaults.

luke-hill:
Hi @eregon this contradicts the advice of @MSP-Greg who I believe is also responsible here

eregon:
Report this to rubygems or bundler then, we are not responsible for those bugs here.
If you want to make a PR to select a better rubygems version for old Rubies like suggested in https://github.com/ruby/setup-ruby/issues/441#issuecomment-1374923260, you are welcome to do so.

eregon:
Also yes, please check locally, I'm pretty sure you'll see the same problem.

eregon:
[...]

PR #551, opened 2023-12-24
Fixup 'rubygems: latest'

MSP-Greg:
Fixes using 'rubygems; latest' for MRI Ruby 2.3 and later.  I tried updating Ruby 2.2 and got errors for two different versions.

`gem update --system` will not allow a bounded version, only an exact version.  So, code is somewhat brittle if new patch releases are done for older RubyGems versions.

Code works same as existing for non MRI Rubies.

One thing I wasn't sure about is whether Ruby master/head builds should not have anything done for 'rubygems; latest'.  Sometime this year I think the RubyGems version was a pre-release, so latest would install an earlier version?  Not.Sure.

Happy Holidays...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -169,15 +169,25 @@ jobs:
       - run: ruby -v
 
   testLatestRubygemsVersion:
-    name: "Test rubygems: latest upgrades the default RubyGems version"
+    name: "Ruby ${{ matrix.ruby }}: latest upgrades RubyGems"
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - { ruby: '3.2', expected_rubygems_version: '3.5.3'  }
+        - { ruby: '3.0', expected_rubygems_version: '3.5.3'  }
+        - { ruby: '2.7', expected_rubygems_version: '3.4.22' }
+        - { ruby: '2.6', expected_rubygems_version: '3.4.22' }
+        - { ruby: '2.5', expected_rubygems_version: '3.3.27' }
+        - { ruby: '2.3', expected_rubygems_version: '3.3.27' }
     steps:
     - uses: actions/checkout@v4
     - uses: ./
       with:
-        ruby-version: '2.6'
+        ruby-version: ${{ matrix.ruby }}
         rubygems: latest
-    - run: ruby -e "exit(Gem.rubygems_version > Gem::Version.new('3.0.3'))"
+    - run: ruby -e 'puts Gem::VERSION; exit(Gem.rubygems_version >= Gem::Version.new("${{ matrix.expected_rubygems_version }}"))'
 
   testFixedRubygemsVersionUpgrades:
     name: "Test rubygems: version upgrades RubyGems to that version if the default is older"
--- a/action.yml
+++ b/action.yml
@@ -12,7 +12,8 @@ inputs:
     description: |
       The version of RubyGems to use. Either 'default' (the default), 'latest', or a version number (e.g., 3.3.5).
       For 'default', no action is taken and the version of RubyGems that comes with Ruby by default is used.
-      For 'latest', `gem update --system` is run to update to the latest RubyGems version.
+      For 'latest', `gem update --system` is run to update to the latest compatible RubyGems version.
+      Ruby head/master builds and Ruby 2.2 and earlier will not be updated.
       Similarly, if a version number is given, `gem update --system <version>` is run to update to that version of RubyGems, as long as that version is newer than the one provided by default.
   bundler:
     description: |
--- a/rubygems.js
+++ b/rubygems.js
@@ -1,8 +1,9 @@
+const common = require('./common')
 const path = require('path')
 const exec = require('@actions/exec')
 const semver = require('semver')
 
-export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix) {
+export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix, platform, engine, version) {
   const gem = path.join(rubyPrefix, 'bin', 'gem')
 
   let gemVersion = ''
@@ -18,7 +19,7 @@ export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix) {
 
   if (rubygemsVersionInput === 'latest') {
     console.log('Updating RubyGems to latest version')
-    await exec.exec(gem, ['update', '--system'])
+    await rubygemsLatest(gem, platform, engine, version)
   } else if (semver.gt(rubygemsVersionInput, gemVersion)) {
     console.log(`Updating RubyGems to ${rubygemsVersionInput}`)
     await exec.exec(gem, ['update', '--system', rubygemsVersionInput])
@@ -27,4 +28,28 @@ export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix) {
   }
 
   return true
+}
+
+// Older RubyGems versions do not account for 'required_ruby_version' when
+// running 'gem update --system', so we have to force a compatible version of
+// rubygems-update.  See https://github.com/ruby/setup-ruby/pull/551 and
+// https://github.com/rubygems/rubygems/issues/7329
+async function rubygemsLatest(gem, platform, engine, version) {
+  if (engine === 'ruby') {
+    const rubyFloatVersion = common.floatVersion(version)
+    if (common.isHeadVersion(version)) {
+      console.log('Ruby master builds use included RubyGems')
+    } else if (rubyFloatVersion >= 3.0) {
+      await exec.exec(gem, ['update', '--system'])
+    } else if (rubyFloatVersion >= 2.6) {
+      await exec.exec(gem, ['update', '--system', '3.4.22'])
+    } else if (rubyFloatVersion >= 2.3) {
+      await exec.exec(gem, ['update', '--system', '3.3.27'])
+    } else {
+      console.log(`Cannot update RubyGems for Ruby version ${version}`)
+    }
+  } else {
+    // non MRI Rubies (TruffleRuby and JRuby)
+    await exec.exec(gem, ['update', '--system'])
+  }
 }
--- a/dist/index.js
+++ b/dist/index.js
@@ -65086,11 +65086,12 @@ __nccwpck_require__.r(__webpack_exports__);
 /* harmony export */ __nccwpck_require__.d(__webpack_exports__, {
 /* harmony export */   "rubygemsUpdate": () => (/* binding */ rubygemsUpdate)
 /* harmony export */ });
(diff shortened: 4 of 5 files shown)


MSP-Greg:
Added a second commit which skips `gem update --system` for Ruby master/head builds.  Added a note to `action.yml`.

MSP-Greg:
JFYI...

Puma separates MRI & non MRI CI.

In MRI CI, we didn't use the rubygems input, as it really didn't work.  We added a step to fix RubyGems/Bundler in older Rubies.  Note that Puma CI tests Bundler functionality, as it's used when restarting/reloading Puma.

I used this PR in Puma CI, and it works as expected.  I removed the additional step needed for older Rubies.

Head builds are left alone, and Rubies are upgraded with the proper version of RubyGems.  We test Ruby 2.4 and later on all platforms.

Happy Holidays, Greg

MSP-Greg:
Note that with the release of RubyGems 3.5.x (restricted to Ruby 3.x), all Ruby versions 2.x fail when the rubygems input is set to latest.

See https://github.com/MSP-Greg/setup-ruby/actions/runs/7330308381

MSP-Greg:
@eregon

Any thoughts on TruffleRuby and maybe JRuby?  Handle like Ruby head builds?

rubygems.js
```diff
@@ -28,3 +28,19 @@ export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix) {
 
   return true
 }
+
+async function rubygemsLatest(gem, platform, engine, version) {
```
eregon, line 33: Could you add a link here to https://github.com/rubygems/rubygems/issues/7329 or another appropriate issue explaining why we need to workaround this bug in RubyGems?

yarn.lock
```diff
@@ -42,11 +42,12 @@
     minimatch "^3.0.4"
 
 "@actions/http-client@^2.0.1", "@actions/http-client@^2.1.1":
-  version "2.1.1"
-  resolved "https://registry.yarnpkg.com/@actions/http-client/-/http-client-2.1.1.tgz#a8e97699c315bed0ecaeaaeb640948470d4586a0"
-  integrity sha512-qhrkRMB40bbbLo7gF+0vu+X+UawOvQQqNAA/5Unx774RS8poaOhThDOG6BGmxvAnxhQnDp2BG/ZUm65xZILTpw==
+  version "2.2.0"
```
eregon, line 45: Could you undo all changes in yarn.lock?
They can cause failures, etc and updating usually makes it worse more than it helps from past experience.
In any case updating it should be done in its own commit/PR so easy to revert if any issue.

eregon:
> Any thoughts on TruffleRuby and maybe JRuby? Handle like Ruby head builds?

TruffleRuby ignores `gem update --system` currently so nothing special to do there.
For JRuby I don't know, but simpler is better.

rubygems.js
```diff
@@ -28,3 +28,19 @@ export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix) {
...
+  if (engine === 'ruby') {
+    if (version >= '3.0') {
```
eregon: Comparing versions as strings feels brittle, could you use common.floatVersion() instead?

rubygems.js
```diff
@@ -28,3 +28,19 @@ export async function rubygemsUpdate(rubygemsVersionInput, rubyPrefix) {
...
+      await exec.exec(gem, ['update', '--system'])
+    } else if (version >= '2.6') {
+      await exec.exec(gem, ['update', '--system', '3.4.22'])
+    } else if (version >= '2.3') {
+      await exec.exec(gem, ['update', '--system', '3.3.27'])
+    } else {
+      console.log(`Cannot update RubyGems for Ruby version ${version}`)
+    }
+  } else {
+    exec.exec(gem, ['update', '--system'])
```
eregon, line 47: Missing `await`

.github/workflows/test.yml
```diff
@@ -169,15 +169,25 @@ jobs:
...
 
   testLatestRubygemsVersion:
-    name: "Test rubygems: latest upgrades the default RubyGems version"
+    name: "Ruby ${{ matrix.ruby }}: latest upgrades RubyGems"
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - { ruby: '3.2', rg_vers: '3.5.3'  }
```
eregon, line 178: rg_vers -> rubygems_version/expected_version/expected_rubygems_version (abbreviations are an overhead on readability if not super well known ones)

.github/workflows/test.yml
```diff
@@ -169,15 +169,25 @@ jobs:
...
+        - { ruby: '2.3', rg_vers: '3.3.27' }
     steps:
     - uses: actions/checkout@v4
     - uses: ./
       with:
-        ruby-version: '2.6'
+        ruby-version: ${{ matrix.ruby }}
         rubygems: latest
-    - run: ruby -e "exit(Gem.rubygems_version > Gem::Version.new('3.0.3'))"
+    - run: ruby -e "puts Gem::VERSION; exit(Gem.rubygems_version >= Gem::Version.new('${{ matrix.rg_vers }}'))"
```
eregon, line 190: Could you swap single and double quotes? It's safer to use single quotes outside for Ruby code (was already an issue before but since you're touching this line let's do it)

action.yml
```diff
@@ -12,7 +12,7 @@ inputs:
     description: |
       The version of RubyGems to use. Either 'default' (the default), 'latest', or a version number (e.g., 3.3.5).
       For 'default', no action is taken and the version of RubyGems that comes with Ruby by default is used.
-      For 'latest', `gem update --system` is run to update to the latest RubyGems version.
+      For 'latest', `gem update --system` is run to update to the proper latest RubyGems version.  Ruby head/master builds and Ruby 2.2 and earlier will not be updated.
```
eregon, line 15: Keep a sentence per line

eregon, line 15: `proper latest RubyGems` -> `latest compatible RubyGems`

eregon:
@MSP-Greg I'm curious, why do you update RubyGems in Puma's CI?
I don't recall any reason to update RubyGems in CI at the moment.

eregon (changes requested):
Looks great but needs some polishing (see comments)

rubygems.js
MSP-Greg: As mentioned, this has always been an issue, but since it's only tested here on Ruby 2.6, it failed because of the recently changed minimum Ruby version in RubyGems.

RubyGems 3.4 supported Ruby 2.6 and later, RubyGems 3.5 supports Ruby 3.0 and later.

yarn.lock
MSP-Greg: I'll split the yarn.lock changes.

rubygems.js
MSP-Greg: Expecting a `x.10` version?  Regardless, I'll update.

rubygems.js
MSP-Greg: Sorry, that line is only for non MRI Ruby.  Given our/my lack of knowledge of JRuby, remove it and log a message that RubyGems is not updated on non MRI platforms?

Nevermind, I'll fix it as you've listed above.  So, same behavior as current with non MRI.

MSP-Greg:
> I'm curious, why do you update RubyGems in Puma's CI?

Honestly, I can't recall. I wrote the code and included a note with `required_ruby_version`.  Older RubyGems/Bundler were very quirky with that, especially when one throws in cross-platform and pre-compiled gems being available.

> but needs some polishing

Not a good phrase to use.  Conversely, I guess I shouldn't write code when I'm also doing 'pre Christmas Eve dinner prep'...

This issue has come up in the past, a few issues related.  In one [message](https://github.com/ruby/setup-ruby/issues/529#issuecomment-1744868928) you said:

> I can't care less about Ruby 2.5 and old RubyGems bugs, that's the problem of anyone still using Ruby 2.5.

I've advocated using newer Ruby versions, and also using head builds in CI, but I also understand testing against older versions.

As you've stated, this is fixing 'old RubyGems bugs'.  I believe current RubyGems/Bundler has fixed all the bugs (thanks to some hard work by very few people).

Although this code may need the RubyGems versions updated from time to time (so it's a burden here), this fix will benefit others that maybe aren't as familiar with all the issues involved with older RubyGems, cross-platform CI and scripting, etc, etc.  For others, it may cleanup there Action workflow scripts...

MSP-Greg:
@eregon I think I fixed/updated all items.

eregon (approved):
Thank you!

eregon:
> Honestly, I can't recall. I wrote the code and included a note with `required_ruby_version`. Older RubyGems/Bundler were very quirky with that, especially when one throws in cross-platform and pre-compiled gems being available.

No worries.

> > but needs some polishing
> 
> Not a good phrase to use. Conversely, I guess I shouldn't write code when I'm also doing 'pre Christmas Eve dinner prep'...

What's wrong with it? I didn't mean anything bad. Is "needs some cleanup" better?
I just meant review should be addressed before merging, because it's not just trivial/stylistic things (which could be cleaned up separately later easily).
But it's fairly obvious so maybe I shouldn't have said anything.

> This issue has come up in the past, a few issues related. In one [message](https://github.com/ruby/setup-ruby/issues/529#issuecomment-1744868928) you said:
> 
> > I can't care less about Ruby 2.5 and old RubyGems bugs, that's the problem of anyone still using Ruby 2.5.

Yeah in general I'm unhappy to have to workaround RubyGems/Bundler bugs, specifically I want to ensure such bugs are properly known upstream and fixed there, but I understand in cases such as this one (where we cannot fix RubyGems shipped with some old Ruby) it's warranted.
I'm still unsure whether updating RubyGems is ever necessary/a good idea or it brings more problems than it helps (that's my current impression based on a few cases of people trying to use it and asking help). But regardless of that I don't want to reopen that debate here and it does seem some people use it.
I appreciate your PR to fix this.

> I've advocated using newer Ruby versions, and also using head builds in CI, but I also understand testing against older versions.

The fact this action supports Ruby 1.9 means we try to support as much as possible, but still stay reasonable and e.g. not accumulate many fixes for old Rubies (e.g. if it doesn't compile anymore, we don't try to patch that Ruby, it's just not available).
IMO if there is something in the ecosystem breaking for ancient Rubies it's a good push to stop supporting that Ruby in gems (BTW I wouldn't consider 2.7 ancient yet).

dentarg:
> I'm curious, why do you update RubyGems in Puma's CI?
> I don't recall any reason to update RubyGems in CI at the moment.

Here's an example from Sinatra CI, on Ruby 3.0 we use `latest` RubyGems version: https://github.com/sinatra/sinatra/blob/5d844eecdc349d32c5c7de72fe68b5ebaafdbee5/.github/workflows/test.yml#L63-L70 due to https://github.com/rubygems/rubygems/pull/6490

PR #567, opened 2024-01-30
Add support for macos-14 GitHub runners

### LINKED ISSUE #568: macos-14-arm64 detected as self-hosted

- [X] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [X] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

GitHub-hosted runner

https://github.com/slackhq/circuit/actions/runs/7716099695/job/21032136995?pr=1165#step:4:9

### Any other notes?

_No response_

ntkme:
This PR adds support for new macos-14 GitHub hosted runners.

Here are all current hosted runners with macos 13 or 14:

- macos-13 (x64)
- macos-13-large (x64)
- macos-13-xlarge (arm64)
- macos-14 (arm64)
- macos-14-large (x64)
- macos-14-xlarge (arm64)

References:

- https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
- https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
- https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-macos-larger-runners

---

- Closes #568


ntkme:
> Thank you for this proposal. It looks fine to me but probably getDownloadURL needs to be revisited as well, see [here](https://github.com/ruby/setup-ruby/blob/bd03e04863f52d169e18a2b190e8fa6b84938215/ruby-builder.js#L112)

That is actually correct. It is the platform of the ruby ci used builds the binary release.

README.md
```diff
@@ -43,7 +43,7 @@ The action works on these [GitHub-hosted runners](https://docs.github.com/en/act
 | Operating System | Supported |
 | ---------------- | --------- |
 | Ubuntu  | `ubuntu-20.04`, `ubuntu-22.04` |
-| macOS   | `macos-11`, `macos-12`, `macos-13`, `macos-13-xlarge` (arm64) |
+| macOS   | `macos-11`, `macos-12`, `macos-13`, `macos-13-xlarge` (arm64), `macos-14` (arm64) |
```
LaStrada, line 46: And shouldn't this include `macos-14-large` and `macos-14-xlarge`? And also `macos-13-large`?

Are we sure `macos-14` is running on ARM? I can see that GitHub writes two different things about the runners.
Here they write that `Large` is Intel, `XLarge` is arm64:
https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-macos-larger-runners

Here they write that `macos-14` is M1:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
```suggestion
| macOS   | `macos-11`, `macos-12`, `macos-13`, `macos-13-large`, `macos-13-xlarge` (arm64), `macos-14`, `macos-14-large` (arm64), `macos-xlarge` (arm64) |
```

ntkme: Please read the description of this PR again and link referenced there.

ntkme: Also, I think `macos-13-xlarge` got a special callout only because it was the first and the only github hosted macos arm64 runner at the time. Typically larger runners are not explicitly listed in the doc.

ntkme: And finally, M1 is arm64. https://en.wikipedia.org/wiki/Apple_M1

LaStrada: Yes you're right, I misunderstood the first link after reading the next ones. And if `macos-13-large` was a "special" one, please ignore my comments!

(but `macos-13-xlarge` should probably be removed now that `macos-14` is out, either right away or at a later point)

ntkme:
We can probably remove the usage of self-hosted `macos-arm-oss` and rely on macos-14 (arm64) for CI, but I've kept it for now.

README.md
ntkme: Updated README to remove `macos-13-xlarge`.

.github/workflows/test.yml
```diff
@@ -33,6 +33,14 @@ jobs:
         exclude:
         # https://github.com/ruby/setup-ruby/issues/496
         - { os: ubuntu-22.04, ruby: '2.2' }
+        # Too old, fails to compile on macos-14 (arm64)
```
eregon, line 36: `fails to compile`? Do you have an example error?
Since we're not compiling these Rubies on macos-14 do you mean C extensions fail to compile?

eregon (approved):
Thank you, this looks great.
I'll let the CI run and then merge.

.github/workflows/test.yml
eregon: Ah it's just the same as `# Too old, fails to compile on macos-arm-oss` below, I'll tweak the comment post-merge

eregon:
> We can probably remove the usage of self-hosted `macos-arm-oss` and rely on macos-14 (arm64) for CI, but I've kept it for now.

It's not self-hosted but macos arm runners provided by GitHub for the Ruby org.
Looking at it, `macos-arm-oss` is macOS 13.6.3 while `macos-14` is macOS 14.2.1.
We need to keep building on macOS 13 to support these: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-macos-larger-runners

PR #584, opened 2024-04-22
Respect $HOME on macOS for caching

### LINKED ISSUE #583: Support non `runner` user on self-hosted runners

- [X] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [X] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

self-hosted runner

https://github.com/cirruslabs/actions-playground/actions/runs/8781520697/job/24093712039

### Any other notes?

Self-mosted macOS runners not always use `runner` as the user name. Right now `runner` user name is hard coded:

https://github.com/ruby/setup-ruby/blob/6bd3d993c602f6b675728ebaecb2b569ff86e99b/common.js#L267-L269

Can probably be fixed similar to https://github.com/actions/setup-python/pull/708

fkorotkov:
Fixes #583

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ if the [Runner Image](https://github.com/actions/runner-images) is very similar 
 * Make sure to use the same operating system and version.
 * Make sure to use the same version of libssl.
 * Make sure that the operating system has `libyaml-0` and [`libgmp`](https://stackoverflow.com/questions/26555902/ruby-v-dyld-library-not-loaded-usr-local-lib-libgmp-10-dylib) installed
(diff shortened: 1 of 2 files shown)


eregon:
Have you tried this?
I'm pretty sure it doesn't work.
The reason is the prebuilt Ruby doesn't support to be moved around, so it needs to be exactly under `/Users/runner/hostedtoolcache`, or it will fail to run.
There is more information about this in previous tickets.

PR #593, opened 2024-05-06
Upstream supported dev dependency versions

benlangfeld:


eregon:
As mentioned in https://github.com/ruby/setup-ruby/pull/592#discussion_r1590278937 we should instead remove those files (and add tests creating these files dynamically ideally, right now these files are unused), could you update the PR to do that?

benlangfeld:
These files are very much used in the local environment of a developer working on this project. I do not think dynamically generating them for CI is at all related to the concern this PR fixes.

eregon:
That's not their purpose in this project, e.g. see https://github.com/ruby/setup-ruby/commits/master/.ruby-version https://github.com/ruby/setup-ruby/commits/master/.tool-versions
They were only added for testing, but probably these tests got removed.

There is no need for any particular version of these tools to work on this project, it's JS anyway.

benlangfeld:
> There is no need for any particular version of these tools to work on this project, it's JS anyway.

There is no need for any particular version of these tools to work on this project, it's JS anyway.

NodeJS and Yarn are required. Is it not preferable for these to be readily available to contributors via a dependency management solution (this or Docker, for example) than to require their manual installation/selection?

eregon:
Look, I know why I added them, the versions there are meant to test setup-ruby picking the ruby version from it, nothing else.
And I do not want to bother updating these or get noise in PRs from them.

PR #594, opened 2024-05-06
Remove trailing whitespace

benlangfeld:
This supposedly forces linebreaks but I don't see any difference in rendered output on github.com. I don't think it's actually serving any purpose.


eregon:
See https://github.com/ruby/setup-ruby/pull/592#discussion_r1591310937

benlangfeld:
I found the difference, but I think this is actually a readability improvement, as well as being compatible with most developers' good practice of auto-trimming trailing whitespace in their editors. I urge you to reconsider.

PR #596, opened 2024-05-14
Support ubuntu-24.04-x64

### LINKED ISSUE #595: Faild in ubuntu-24.04 runner
Ubuntu 24.04 runner has been released. 

https://github.com/actions/runner-images/issues/9848

But, attempting to use it resulted in the following errors.

```
Error: The current runner (ubuntu-24.04-x64) was detected as self-hosted because the platform does not match a GitHub-hosted runner image (or that image is deprecated and no longer supported).
In such a case, you should install Ruby in the $RUNNER_TOOL_CACHE yourself, for example using https://github.com/rbenv/ruby-build
You can take inspiration from this workflow for more details: https://github.com/ruby/ruby-builder/blob/master/.github/workflows/build.yml
$ ruby-build head /opt/hostedtoolcache/Ruby/head/x64
Once that completes successfully, mark it as complete with:
$ touch /opt/hostedtoolcache/Ruby/head/x64.complete
It is your responsibility to ensure installing Ruby like that is not done in parallel.
```

kachick:
Resolves #595

As I understand it, https://github.com/ruby/ruby-builder/pull/17 is needed as a preparation task

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, ubuntu-22.04, macos-11, macos-12, macos-13, macos-14, macos-arm-oss, windows-2019, windows-2022 ]
+        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04, macos-11, macos-12, macos-13, macos-14, macos-arm-oss, windows-2019, windows-2022 ]
         ruby: [
           '1.9', '2.0', '2.1', '2.2', '2.3', '2.4', '2.5', '2.6', '2.7', '3.0', '3.1', '3.2', '3.3', ruby-head,
           jruby, jruby-head,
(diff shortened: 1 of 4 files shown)


eregon:
Could you also list 24.04 in the README?

.github/workflows/test.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         exclude:
```
eregon, line 33: Could you exclude 1.9 on 24.04? (see linked issue)
I scheduled a second run on ruby-builder but I guess it will fail too.

.github/workflows/test.yml
```diff
@@ -33,6 +33,7 @@ jobs:
...
         # https://github.com/ruby/setup-ruby/issues/496
         - { os: ubuntu-22.04, ruby: '2.2' }
+        # - { os: ubuntu-24.04, ruby: '2.2' }
```
eregon, line 36: Could you either remove this line if unneeded or if the build does fail then uncomment it?

eregon:
CI should be close to passing on the next run, but we still need to update the 3 -dev-builder repos.

eregon:
For ruby-head and jruby-head we have a cyclic dependency:
https://github.com/ruby/ruby-dev-builder/actions/runs/9148670190/job/25151463375
https://github.com/ruby/jruby-dev-builder/actions/runs/9148670127/job/25151463255
So we'll have to merge those as red or exclude them, I think merging as red is fine for this case.

eregon (approved):

PR #597, opened 2024-05-15
Add support for ubuntu 24.04

toadjaune:
Hi folks !

Following the release of the ubuntu 24.04 image by github actions, adding support is needed here, this PR attempts to add it.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -525,6 +525,7 @@ async function hashFile(file) {
 const GitHubHostedPlatforms = [
   'ubuntu-20.04-x64',
   'ubuntu-22.04-x64',
+  'ubuntu-24.04-x64',
   'macos-11-x64',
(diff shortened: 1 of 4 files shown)


toadjaune:
Oh, I'm stupid, this is a duplicate of #596

Sorry about that :/

dist/index.js
```diff
@@ -65937,4 +65938,4 @@ if (__filename.endsWith('index.js')) { run() }
 
 module.exports = __webpack_exports__;
 /******/ })()
-;
\ No newline at end of file
+;
```
eregon, line 65941: There is an extra newline here which will most likely cause CI to fail.

Anyway I'll use https://github.com/ruby/setup-ruby/pull/596 since that was created earlier and both PRs are basically the same.
Thank you for the PR though.

PR #613, opened 2024-06-26
[ubicloud] Use ubicloud/cache instead of actions/cache in the workflow

enescakir:
We just switched the cache package in the package.json file from "actions/cache" to "ubicloud/cache".

When we execute `npm run package`, it simply puts the dependency packages to the dist directory. This is why the change appears less clean.


enescakir:
Wrong base branch, sorry

Issue #650, opened 2024-10-01
No "available interpreter-os association" is visible on the action?


stdedos:
i.e. on https://github.com/ruby/setup-ruby/blob/f321cf5a4d1533575411f8752cf25b86478b0442/ruby-builder-versions.json I don't see which interpreter is available on which OS.

e.g. even though `truffleruby+graalvm-21` supposedly exists, it does not work in my action (https://github.com/stdedos/interception/actions/runs/11123894333/job/30908313355)

- Should you create a matrix? (ie like https://github.com/actions/setup-python/issues/15#issuecomment-525410806)
- (For both repositories), should that be (a) automatically done, and (b) give some "permalink-like" url to "that matrix"?

eregon:
In general it's every version is available on all platforms on which it can be built.
The full list is at https://github.com/ruby/ruby-builder/releases/tag/toolcache FWIW.
There are some known version-platform combinations which do not work: https://github.com/ruby/setup-ruby/blob/c04af2bb7258bb6a03df1d3c1865998ac9390972/index.js#L173-L195

Regarding `truffleruby+graalvm-21` that's just really old and I didn't build it on newer runners.
I would suggest to only test the latest release of truffleruby (with just `ruby-version: truffleruby`).
`truffleruby+graalvm` is basically redundant in CI (unless you need interop with Java or other languages).

Maintaining a matrix would be too high overhead.

stdedos:
> The full list is at [ruby/ruby-builder@`toolcache` (release)](https://github.com/ruby/ruby-builder/releases/tag/toolcache) FWIW.

That is amazing already 😁 

Would you consider mentioning it here? https://github.com/ruby/setup-ruby/blob/c04af2bb7258bb6a03df1d3c1865998ac9390972/README.md?plain=1#L39-L57

> The combinations of the available versions and runners can be seen here: https://github.com/ruby/ruby-builder/releases/tag/toolcache

(_tl;dr that's what I'm asking here_)

eregon:
Ah yes, sure, PR welcome

PR #651, opened 2024-10-02
`README.md`: Mention link to supported Ruby versions for GitHub-hosted runners

stdedos:
Fixes https://github.com/ruby/setup-ruby/issues/650

Signed-off-by: Stavros Ntentos <133706+stdedos@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ This action currently supports these versions of MRI, JRuby and TruffleRuby:
 | `truffleruby` | 19.3.0 - 24.1.0, head |
 | `truffleruby+graalvm` | 21.2.0 - 24.1.0, head |
 
+The combinations of the available versions and runners can be seen [here](https://github.com/ruby/ruby-builder/releases/tag/toolcache).
+
 `ruby-debug` is the same as `ruby-head` but with assertions enabled (`-DRUBY_DEBUG=1`).
 
 Regarding Windows ruby master builds, `mingw` is a MSYS2/MinGW build, `head` & `ucrt` are MSYS2/UCRT64


eregon (approved):
Thanks

stdedos:
> Thanks

... I would appreciate a heads up before merging 😕

I won't try to source an "very authoritative source", but `[here]()` urls can be a very bad thing to do from a11y point-of-view, e.g. https://heyoka.medium.com/dont-use-click-here-f32f445d1021#a1f6

eregon:
The entire README has no URLs as plain text, because that takes a very large amount of space and IMO hurts readability. So I did the obvious fix, keeping your sentence as-is, and I undid the unrelated whitespace changes which are intentional.

Feel free to propose a better wording or linking.

stdedos:
> and I undid the unrelated whitespace changes (the spaces are intentional and have an effect).

Those changes were _unintentional_ from my part. All of my smart editors are configured to prune trailing whitespace.

What effect do they have on Github Commonmark? 😕

Note that https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#paragraphs mentions

>You can create a new paragraph by leaving a blank line between lines of text.

and there is a testament of the GhFM-feature on the internet: https://stackoverflow.com/a/24576087/2309247

It is mentioned in https://github.github.com/gfm/#hard-line-break, https://docs.gitlab.com/ee/user/markdown.html#newlines for ways to force a line break (including the N>1 trailing spaces) - but even if https://stackoverflow.com/a/24576087/2309247 is not mentioned, it still happens.
See this comment for example. 😅

eregon:
It forces a newline, without extra vertical spacing, as e.g. in https://github.com/ruby/setup-ruby?tab=readme-ov-file#caching-bundle-install-automatically
It doesn't affect comments but it does affect the README.

PR #653, opened 2024-10-03
Add ASAN CRuby build

ioquatix:
Expose this build: https://github.com/ruby/ruby-dev-builder/blob/19d43276e3490260c6785742d5c5b93858149f6d/.github/workflows/build.yml#L138-L167

Example usage: https://github.com/socketry/io-event/pull/115


dentarg:
> Expose this build: https://github.com/ruby/ruby-dev-builder/blob/19d43276e3490260c6785742d5c5b93858149f6d/.github/workflows/build.yml#L138-L167

Just saving other people some clicks, that's "Ruby with ASAN", more context in https://github.com/ruby/ruby-dev-builder/pull/10

dentarg:
@ioquatix From https://github.com/DataDog/setup-ruby/commit/0c7206d6db81faf999795ceebfac00d164298bd5 (found in https://github.com/DataDog/dd-trace-rb/pull/3864) it looks like some more changes are needed?

ioquatix:
@dentarg thanks for the links and feedback - @ivoanjo thanks for already investigating this!

ioquatix:
Seems to be building okay: https://github.com/ruby/setup-ruby/actions/runs/11171164332/job/31055440201?pr=653

eregon:
@KJTsanaktsidis are those builds ready to be publicized?

One issue is that build is marked as optional since https://github.com/ruby/ruby-dev-builder/commit/19d43276e3490260c6785742d5c5b93858149f6d as it was failing for a while.

Also there seems to be some issues with ruby-head since a few days: https://github.com/ruby/ruby-dev-builder/actions/runs/11168122071/job/31045848723

KJTsanaktsidis:
Maybe it might be best to give me a week or so to try and get everything back into shape here, I haven't touched any of this ASAN stuff for a little while and I should probably get back into it before we pull the trigger perhaps?

KJTsanaktsidis:
On reflection - I think i'm happy to merge this I guess, the only way to shake out the kinks is if a few brave souls give it a spin. Maybe let's soft-launch this without a README update since Samuel has a use-case for it now, and we can work through any issues and come back in a little while and add an entry to the README.

ivoanjo:
Thanks y'all for looking into this! We had to disable asan testing on https://github.com/DataDog/dd-trace-rb/pull/3915 since the builds were broken for a few days, but I'm happy to work with the fact they are considered experimental and these kinds of things may show up from time to time -- I think they provide enough value as-is ;)

eregon:
I have reverted https://github.com/ruby/ruby-dev-builder/commit/19d43276e3490260c6785742d5c5b93858149f6d to try to always have builds with ASAN, if there are problems for a few days I might make it optional again.

I would also like to get all ruby-dev-builder jobs fixed before merging this.

eregon (approved):

ioquatix:
Thanks everyone! Super excited to try this out!

ioquatix:
Looks like it's working quite okay: https://github.com/socketry/io-event/actions/runs/11196675213/job/31125798429?pr=115

eregon:
@ioquatix Could you create a PR to document this in the README? That way we don't forget about it and we can merge it after a bit more testing.

ioquatix:
Sure

eregon:
I noticed https://github.com/DataDog/dd-trace-rb/actions/runs/11214528450/job/31169787558 so that seems to report some leaks in the Prism compiler in CRuby e.g.
```
 Direct leak of 1232 byte(s) in 22 object(s) allocated from:
    #0 0x5630b0a263dd in calloc (/home/runner/.rubies/ruby-asan/bin/ruby+0xc73dd) (BuildId: d4c818bb4e8ac5804a6050a3d1ebcb0a3ad54afe)
    #1 0x7f83566ee6fa in pm_node_alloc /home/runner/work/ruby-dev-builder/ruby-dev-builder/prism/prism.c:1913:20
    #2 0x7f83566ee6fa in pm_program_node_create /home/runner/work/ruby-dev-builder/ruby-dev-builder/prism/prism.c:6382:31
    #3 0x7f83566ee6fa in parse_program /home/runner/work/ruby-dev-builder/ruby-dev-builder/prism/prism.c:21970:26
    #4 0x7f83566ee6fa in pm_parse /home/runner/work/ruby-dev-builder/ruby-dev-builder/prism/prism.c:22364:12
    #5 0x7f835604a478 in pm_parse_string /home/runner/work/ruby-dev-builder/ruby-dev-builder/./prism_compile.c:11013:23
    #6 0x7f835629d962 in pm_iseq_compile_with_option /home/runner/work/ruby-dev-builder/ruby-dev-builder/iseq.c:1312:17
    #7 0x7f835629d962 in iseqw_s_compile_parser /home/runner/work/ruby-dev-builder/ruby-dev-builder/iseq.c:1520:16
```

I don't know if it's accurate or maybe some false positives, but either way @ivoanjo could you report that to https://github.com/ruby/prism/issues?

eregon:
It's interesting @ioquatix's build in https://github.com/socketry/io-event/actions/runs/11196675213/job/31125798429?pr=115 didn't see those.
Maybe the difference is @ivoanjo's job sets `RUBY_FREE_AT_EXIT=1`, as well as `LSAN_OPTIONS=verbosity=0:log_threads=1:suppressions=`pwd`/suppressions/lsan.supp ASAN_OPTIONS=detect_leaks=1`.
Should we maybe always set (some of?) that for asan builds @KJTsanaktsidis ? Or document it in #654?

ioquatix:
I think adding documentation is a good idea.

ivoanjo:
Well spotted indeed! I saw the new prism complaints, but there were failures on a few of our specs caused by latest rubygems/bundler so I decided to tackle those first before circling back to the prism stuff, and then got sidetracked a bit.

In particular, maybe prism can also be added to suppressions for now.

eregon:
I guess this kind of goes back to adding ASAN in ruby/ruby CI, I think @KJTsanaktsidis was working on that.
That should find these Prism compiler cases for instance.
I think it's probably best to not publicize ASAN builds until that's done, as otherwise it'll be basically guaranteed to have ASAN errors which need to be solved in CRuby.

Adding valgrind in ruby/ruby CI would also be good (even if only running Prism tests or some small workload it would still find many things), I suggested that to @kddnewton since the `memcheck` job in ruby/prism has been finding lots of CRuby leaks (unrelated to Prism).

ioquatix:
I want to offer a counter point: ASAN is enabled in `io-event` gem and I've found bugs using it (including issues in CRuby itself). It has been extremely useful, even in it's current form.

KJTsanaktsidis:
I don’t have a heap to add here other that I’m sorry I haven’t had much time to work on ASAN stuff this week - it _should_ be added to Ruby CI and I do want to do that.

> ASAN_OPTIONS=detect_leaks=1

I will say I think this is not the default: https://github.com/ruby/ruby/blob/c43be94f76982d3ffa2ecd28d34172600b81ca31/main.c#L70

I haven’t tackled Asan detected leaks yet (although Peter probably already found most of them lol)

eregon:
I see, so by default there won't be extra leaks coming from CRuby, so I think that's fine then and OK to document the new build now in the README.

kddnewton:
Sorry just to confirm are there currently leaks being reported in the Prism compiler or no?

eregon:
There were some detected in https://github.com/ruby/setup-ruby/pull/653#issuecomment-2404700593 and I assume are still there, but I don't know if they are false positives or real leaks.
You'll need this new `ruby-asan` build (or CRuby built with ASAN) and `ASAN_OPTIONS=detect_leaks=1` to repro and maybe other env vars set in that CI job.

eregon:
I managed to repro at https://github.com/ruby/prism/pull/3182

PR #654, opened 2024-10-06
Add some details about using asan build.

ioquatix:
We could add links to examples and Ruby's documentation but I thought let's keep it short to start with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ This action currently supports these versions of MRI, JRuby and TruffleRuby:
 
 `ruby-debug` is the same as `ruby-head` but with assertions enabled (`-DRUBY_DEBUG=1`).
 
+`ruby-asan` is the same as `ruby-head` but with AddressSanitizer (ASan) enabled, helpful for finding memory issues in native extensions.
+Native extensions are automatically compiled with AddressSanitizer when using `ruby-asan`.
+`ruby-asan` is currently only available on `ubuntu-24.04`.
+
 Regarding Windows ruby master builds, `mingw` is a MSYS2/MinGW build, `head` & `ucrt` are MSYS2/UCRT64
 builds, and `mswin` is a MSVC/VS 2022 build.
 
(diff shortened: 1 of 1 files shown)


dentarg:
The `asan` build will always be built from `head`, right? Something to mention in the README?

README.md
```diff
@@ -220,6 +220,18 @@ It is also possible to cache gems manually, but this is not recommended because
 There are many concerns which means using `actions/cache` is never enough for caching gems (e.g., incomplete cache key, cleaning old gems when restoring from another key, correctly hashing the lockfile if not checked in, OS versions, ABI compatibility for `ruby-head`, etc).
 So, please use `bundler-cache: true` instead and report any issue.
 
+### ASan Debugging
+
+Ruby can be built with AddressSanitizer (ASan) for debugging memory issues. You can use the `asan` variant of Ruby for this:
```
eregon, line 225: Could you move it around lines 23? That's where we explain other similar variants.
There is no need for a YAML snippet, it should be obvious.
It's also ubuntu-24.04 only, we should mention that.

eregon: Mmh, maybe we need its own section to document env vars and stuff, see https://github.com/ruby/setup-ruby/pull/653#issuecomment-2404707671
But anyway I think we should mention `asan` near line 23 and that it's ubuntu-24.04 only

ioquatix: Yes, think it needs it's own section, but we could also add a note on line 23.

README.md
```diff
@@ -220,6 +220,18 @@ It is also possible to cache gems manually, but this is not recommended because
...
+
+```yaml
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: asan
+```
+
+You should also compile your own code with ASan to get the full benefits of this feature.
```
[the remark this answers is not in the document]
KJTsanaktsidis: It will be automatic for extensions yes. If you’re linking in some other library maybe you want to compile that with ASAN too but that’s probably niche enough to to mention here?

KJTsanaktsidis (approved):

PR #672, opened 2024-11-20
README.md - add `macos-15` to supported runner/os list

MSP-Greg:


eregon:
Sorry, I missed that and did https://github.com/ruby/setup-ruby/commit/9a9fe58c47b8bca3b707c7a46d0c67c4a3fe6d89 instead. Not sure which is better, my commit has the advantage it says any higher version is already supported.

MSP-Greg:
No problem.  Given that the other runners are listed individually, not sure about it.  Some people who are not strong English speakers may be confused.  Also, there may be times when there is a bit of delay between new OS runners release and setup-ruby support of them.  Lastly, possible issues when GHA has beta releases.

Not that critical an issue...

eregon:
Right, https://github.com/ruby/setup-ruby/commit/6825c765eb33355054208d865dd846da7da6584f should clarify

PR #688, opened 2025-01-16
Time sensitive: upgrade `@actions/cache` to `^4.0.0`

Link-:
The cache backend service has been rewritten from the ground up for improved performance and reliability. The [@actions/cache](https://github.com/actions/toolkit/tree/main/packages/cache) package now integrates with the new cache service (v2) APIs.

The new service will gradually roll out as of **February 1st, 2025**. The legacy service will also be sunset on the same date. Changes in this release are **fully backward compatible**.

If you do not upgrade, all workflow runs using any of the deprecated [@actions/cache](https://github.com/actions/toolkit/tree/main/packages/cache) packages will fail.

Upgrading to the recommended version should not break or require any changes to your workflows beyond the changes in this PR.

Please cut a new release after merging this PR.

More details: https://github.com/actions/toolkit/discussions/1890

> [!TIP]
> I'd highly recommend adding the configuration below to your dependabot.yml to receive further upgrades we will be making to the cache package

```yml
diff --git a/.github/dependabot.yml b/.github/dependabot.yml
index b18fd29..025b9c1 100644
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,11 @@ updates:
     directory: '/'
     schedule:
       interval: 'weekly'
+
+  # Enable version updates for npm
+  - package-ecosystem: 'npm'
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: '/'
+    # Check the npm registry for updates every day (weekdays)
+    schedule:
+      interval: 'weekly'
\ No newline at end of file
```
---

The tests I ran are positive & everything is working as expected with the new service.

cs_2025-01-16_14-14-30

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "homepage": "https://github.com/ruby/setup-ruby",
   "dependencies": {
-    "@actions/cache": "^3",
+    "@actions/cache": "^4.0.0",
     "@actions/core": "^1.10.0",
     "@actions/exec": "^1.1.1",
     "@actions/io": "^1.1.2",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -336,6 +336,8 @@ jobs:
           ruby-version: '2.6'
           bundler-cache: true
       - run: bundle exec rails --version
+        env:
+          RUBYOPT: '-rlogger' # Rails 6 is missing a require 'logger' in active_support/logger_thread_safe_level.rb
 
   testTruffleRubyNokogiri:
     name: "Test installing a Gemfile with nokogiri on TruffleRuby"
--- a/dist/index.js
+++ b/dist/index.js
@@ -755,11 +755,14 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
 };
 Object.defineProperty(exports, "__esModule", ({ value: true }));
 exports.saveCache = exports.restoreCache = exports.isFeatureAvailable = exports.ReserveCacheError = exports.ValidationError = void 0;
-const core = __importStar(__nccwpck_require__(2186));
+const core = __importStar(__nccwpck_require__(448));
 const path = __importStar(__nccwpck_require__(1017));
 const utils = __importStar(__nccwpck_require__(1518));
 const cacheHttpClient = __importStar(__nccwpck_require__(8245));
+const cacheTwirpClient = __importStar(__nccwpck_require__(2502));
+const config_1 = __nccwpck_require__(5147);
 const tar_1 = __nccwpck_require__(6490);
+const constants_1 = __nccwpck_require__(8840);
 class ValidationError extends Error {
     constructor(message) {
         super(message);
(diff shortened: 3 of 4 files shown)


package-lock.json
```diff
@@ -0,0 +1,905 @@
+{
```
eregon, line 1: Is there any reason to commit this file?
It seems redundant with `yarn.lock`.

package.json
```diff
@@ -33,4 +33,4 @@
   "devDependencies": {
     "@vercel/ncc": "^0.33.4"
   }
-}
+}
```
eregon, line 36: Could you preserve this final newline?

eregon (commented):
Thank you for the PR!

package-lock.json
Link-: No reason, good catch!

package.json
Link-: Adjusted, thank you!

eregon (approved):

MSP-Greg:
@eregon

The source repro for many of the packages in `yarn.lock` changed from `registry.yarnpkg.com` to `https://registry.npmjs.org`.

Is that desired?  Regardless, maybe a `yarn upgrade` is needed?

eregon:
Mmh, I just ran the pre-commit so `yarn install` locally, odd, with `yarn 1.22.21`.
I guess it's fine.
I'd rather avoid touching any other package in this PR to make it easier to assess if anything would break.

`@actions/cache@^4.0.0` has many dependencies including things like `prettier`, but not something we can change here.

eregon:
@MSP-Greg Thanks for noticing, I fixed that now by restoring the yarn.lock from master and rerunning `yarn install`.

eregon:
@Link- Regarding
> I'd highly recommend adding the configuration below to your dependabot.yml to receive further upgrades we will be making to the cache package

Is there a way to only create PRs for updates of the actions/cache package?
Because given the huge amount of transitive dependencies I wouldn't want a PR every week to update some package I don't really care about (unproductive busy work).
Just dealing with security issues used to be a pain when we depended on axios, it was terrible, so I got rid of that dependency.
In my experience so far with JS stuff, auto updates break more things than they solve, so they don't seem worth it.
If there is a known bug fix wanted we can always update something specific.

Link-:
> @Link- Regarding
> 
> > I'd highly recommend adding the configuration below to your dependabot.yml to receive further upgrades we will be making to the cache package
> 
> Is there a way to only create PRs for updates of the actions/cache package? Because given the huge amount of transitive dependencies I wouldn't want a PR every week to update some package I don't really care about (unproductive busy work). Just dealing with security issues used to be a pain when we depended on axios, it was terrible, so I got rid of that dependency. In my experience so far with JS stuff, auto updates break more things than they solve, so they don't seem worth it. If there is a known bug fix wanted we can always update something specific.

No problem at all, it's really up to you how you'd like to manage this. It is possible to use: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#allowing-specific-dependencies-to-be-updated

```
    ignore:
      - dependency-name: "*" # First ignore all dependencies
    allow:
      - dependency-name: "@actions/cache" # Then allow only the specific dependency you want to update
```

Issue #577, opened 2024-03-15
Linux ARM support


sandstrom:
Now that Github Actions has ARM-runners ([in private beta currently](https://github.com/github/roadmap/issues/836)), it would be awesome if this action would support ARM too.

Right now, it seems to assume Github doesn't have ARM.

https://github.com/ruby/setup-ruby/blob/master/common.js#L165

Here is a signup link for the private beta:

https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions/

dentarg:
What's needed to support it? Seems tricky to support it when the docs for it are still private?

sandstrom:
@dentarg If you would like to, I'm sure @lkfortuna (PM working with Github Actions) can get you into the beta, since this action is widely used.

dentarg:
I was thinking in general, if you have access to the beta, you should be able to submit a PR to this repo to show what needs to be done?

I've been invited to the beta, I actually just emailed her earlier today ;-)

sandstrom:
@dentarg Ah, got it!

I would love to help out, but I don't know the details of this action well enough, sorry.

eregon:
It's the same as for any other new runner, we need access to the runner from the ruby organization to build binaries, see https://github.com/ruby/setup-ruby/pull/494#issuecomment-1743537651
Until then linux-aarch64 will be detected as self-hosted, because there are no prebuilt Rubies for it available.

So @lkfortuna @Steve-Glass could you make these new beta linux-aarch64 runners available for the `ruby` organization?
What would be the `runs-on` value?

sandstrom:
@dentarg Yeah, I know 😄 

I've learned a lot by doing things, just that I haven't yet spent that much time with JS-based Github Actions.

I do contribute to a few open-source projects, but nobody can do everything so I'm trying to pick a few. My current side-project is to improve the `ipaddress` gem.

Out of curiosity, do you live in the Stockholm area, or elsewhere? 🏙️

Steve-Glass:
@eregon - Thanks for reaching out. Let me bring this feedback to the team and we will get back to you next week.

sandstrom:
@lkfortuna @Steve-Glass Somewhat off topic, but is there a reason why the `setup-python` and `setup-node` actions are managed by Github, but this action lives outside?

eregon:
[...]

PR #690, opened 2025-01-16
Add arm64 os variants

xrl:
Work towards #577 and based on documentation from [the arm64 public release announcement](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/#how-to-use-the-runners)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04, macos-13, macos-14, macos-15, windows-2019, windows-2022 ]
+        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-22.04-arm, ubuntu-24.04, ubuntu-24.04-arm, macos-13, macos-14, macos-15, windows-2019, windows-2022 ]
         ruby: [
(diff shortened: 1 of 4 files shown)


xrl:
I added to the allow list of platforms and now I'm hitting some odd behavior inside the github actions environment, specifically there's a `ruby` but no `bundle`. and the gem home appears to be superuser only.

```
runner@fv-az1504-216:~/work/setup-ruby/setup-ruby$ gem install json -v 2.2.0
Fetching json-2.2.0.gem
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /var/lib/gems/3.0.0 directory.
```

and

```
runner@fv-az1504-216:~/work/setup-ruby/setup-ruby$ bundle install
Command 'bundle' not found, but can be installed with:
snap install ruby          # version 3.4.1, or
apt  install ruby-bundler  # version 2.3.5-2
See 'snap info ruby' for additional versions.
```

eregon:
Right, that's just the system ruby, so probably there are no Rubies in the hostedtoolcache currently (to be double-checked), since https://github.com/actions/runner-images/blob/main/images/ubuntu/scripts/build/install-ruby.sh uses builds from ruby-builder, and we haven't any for linux-arm64 yet.

I think I will just make my own PR when the other parts are ready, the changes in setup-ruby are not the hard part (building tons of rubies on this new images is), but also more is needed than what's in this PR, and it needs to be coordinated with building rubies, e.g. some older versions likely don't support linux-aarch64.
IOW, it's more work for me to integrate this PR than making my own, so let me close it.

PR #691, opened 2025-01-17
yarn.lock - Remove @actions/core 1.10.1, use @actions/core 1.11.1

MSP-Greg:
With the update `"@actions/cache": "^4.0.0"`, along with the dependency on `"@actions/core": "^1.10.0"`, two versions of @actions/core are listed in `yarn.lock`.

Update `package.json` to use `"@actions/core": "^1.11.1"`.

JFYI, this started while I was working on adding `windows-2025`.  Also, haven't looked at it yet, but 1.11.1 adds a `platform` object, see:

https://github.com/actions/toolkit/tree/main/packages/core#platform-helper

We may be able to use `platform` in the code here...

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -755,7 +755,7 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
 };
 Object.defineProperty(exports, "__esModule", ({ value: true }));
 exports.saveCache = exports.restoreCache = exports.isFeatureAvailable = exports.ReserveCacheError = exports.ValidationError = void 0;
-const core = __importStar(__nccwpck_require__(448));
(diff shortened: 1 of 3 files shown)


eregon (approved):
Thanks!

eregon:
> https://github.com/actions/toolkit/tree/main/packages/core#platform-helper
> 
> We may be able to use `platform` in the code here...

platform.getDetails() might be interesting, indeed.

eregon:
Yep, that seems convenient, I tried it in https://github.com/ruby/setup-ruby/actions/runs/12835965709/job/35796665896#step:3:7
The Windows version is a bit less consistent (part of name), but not hard.

PR #693, opened 2025-01-17
Support new windows-2025 GHA image

MSP-Greg:


eregon (approved):
Nice and simple, thanks!

PR #700, opened 2025-02-04
Add support for mise.toml

hobbypunk90:


index.js
```diff
@@ -130,6 +132,11 @@ function parseRubyEngineAndVersion(rubyVersion) {
     const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s/.test(e))[0]
     rubyVersion = rubyLine.match(/^ruby\s+(.+)$/)[1]
     console.log(`Using ${rubyVersion} as input from file .tool-versions`)
+  } else if (rubyVersion === 'mise.toml') { // Read from mise.toml
```
eregon, line 135: https://mise.jdx.dev/configuration.html has 6 possible locations :sweat_smile: 
Hopefully `mise.toml` is the most common one, I don't really want to logic to search all those.
So yeah, let's look for just `mise.toml` for now.

index.js
```diff
@@ -130,6 +132,11 @@ function parseRubyEngineAndVersion(rubyVersion) {
...
+    const toolVersions = fs.readFileSync('mise.toml', 'utf8').trim()
+    const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s=\s/.test(e))[0]
+    rubyVersion = rubyLine.match(/^ruby\s=\s"(.+)"$/)[1]
```
eregon, line 138: Looking at https://mise.jdx.dev/configuration.html it's also frequent to single-quote the version, could you allow both `'` and `"` in this regexp?

eregon (commented):
Thank you for the PR, it looks good, just one small thing to address

index.js
hobbypunk90: i see, you are right :sweat_smile: especially the conf.d would be crap :grin:

index.js
```diff
@@ -130,6 +132,11 @@ function parseRubyEngineAndVersion(rubyVersion) {
     const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s/.test(e))[0]
     rubyVersion = rubyLine.match(/^ruby\s+(.+)$/)[1]
     console.log(`Using ${rubyVersion} as input from file .tool-versions`)
+  } else if (rubyVersion === 'mise.toml') { // Read from mise.toml
+    const toolVersions = fs.readFileSync('mise.toml', 'utf8').trim()
+    const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s=\s/.test(e))[0]
```
eregon, line 137:
```suggestion
    const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s*=\s*/.test(e))[0]
```

index.js
```diff
@@ -130,6 +132,11 @@ function parseRubyEngineAndVersion(rubyVersion) {
...
+    rubyVersion = rubyLine.match(/^ruby\s=\s['"](.+)['"]$/)[1]
```
eregon, line 138:
```suggestion
    rubyVersion = rubyLine.match(/^ruby\s*=\s*['"](.+)['"]$/)[1]
```

PR #713, opened 2025-02-24
Expose `3.4-asan` Ruby builds from ruby-dev-builder

### LINKED ISSUE #682: Add a stable ruby-asan variant?
Hey :wave:

I've been meaning to ask this once Ruby 3.4 is out. Would it be possible to add a stable ruby-asan variant? (e.g. using stable 3.4 releases?).

My use-case is that we're using ruby-asan [to test the `datadog` gem](https://github.com/DataDog/dd-trace-rb/blob/master/.github/workflows/test-memory-leaks.yaml) and we've been "bitten" a few times with our CI check starting to fail due to unrelated ruby-head changes.

This is annoying because it leaves us in a situation where this check is brittle, but we still don't want to disable it since it may catch issues.

Having a stable ruby-asan would be a really good middle ground, since we'd know that if it went from green to red, it was because of a change we made, not because ruby-head is, as it should be, sometimes bumpy ;)

ivoanjo:
**What does this PR do?**

This PR provides the final piece of the puzzle to fix #682: now that ruby-dev-builder is building `3.4-asan` Rubies since https://github.com/ruby/ruby-dev-builder/pull/15, we can make them available for use.

**Motivation:**

At Datadog, we're using ASan builds to check for issues [in our library](https://github.com/datadog/dd-trace-rb), and having a `3.4-asan` build improves the experience of having these checks be a required CI step. Using the ASan builds built from ruby-head means our CI could break because of unrelated issues/changes from ruby-head.

**Additional Notes:**

`yarn` was trying really hard to update a bunch of things, which generated A LOT of diff noise. I've manually pared them down just to the actual changes related to adding the new variant so the diff makes a bit more sense.

**How to test the change?**

I've already tested this in
https://github.com/DataDog/dd-trace-rb/actions/runs/13392865620/job/37696781848 and it seems to be working fine!


eregon (approved):

eregon:
> `yarn` was trying really hard to update a bunch of things, which generated A LOT of diff noise. I've manually pared them down just to the actual changes related to adding the new variant so the diff makes a bit more sense.

Mmh, that's weird, did that happen with just `yarn install` as in https://github.com/ruby/setup-ruby/blob/master/pre-commit ?

eregon:
We could probably try to use npm again instead of yarn (we used to hit a npm bug but probably no longer the case), not sure it would behave better though.

PR #717, opened 2025-03-04
[Fixes #716] deal with multiple versions in .tool-versions

### LINKED ISSUE #716: Support `.tool-versions` files with multiple versions
ASDF supports having multiple different versions for each dependency.
```
ruby 3.1.6 3.2.4 3.3.3
```
From their docs:
Image

Source: https://asdf-vm.com/manage/configuration.html#tool-versions

This is very useful for gems that have to support multiple different versions of ruby.

Unfortunately `setup-ruby` crashes out when encountering this.

Image

Please add support for this feature of ASDF.

jtannas:
See #716  for explanation

Tested on one of my projects: https://github.com/jtannas/katachi/actions/runs/13660430691/job/38190080032?pr=19

While it might be cool for it to attempt each ruby version in order (aka. mimic ASDF parsing) or to treat multiple `.tool-versions` entries as a matrix, that's way beyond what I need and am comfortable tackling as a first PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,7 +130,7 @@ function parseRubyEngineAndVersion(rubyVersion) {
   } else if (rubyVersion === '.tool-versions') { // Read from .tool-versions
     const toolVersions = fs.readFileSync('.tool-versions', 'utf8').trim()
     const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s/.test(e))[0]
-    rubyVersion = rubyLine.match(/^ruby\s+(.+)$/)[1]
+    rubyVersion = rubyLine.split(/\s/)[1]
     console.log(`Using ${rubyVersion} as input from file .tool-versions`)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -130,7 +130,7 @@ function parseRubyEngineAndVersion(rubyVersion) {
   } else if (rubyVersion === '.tool-versions') { // Read from .tool-versions
     const toolVersions = fs.readFileSync('.tool-versions', 'utf8').trim()
     const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s/.test(e))[0]
-    rubyVersion = rubyLine.match(/^ruby\s+(.+)$/)[1]
+    rubyVersion = rubyLine.split(/\s/)[1]
```
eregon, line 133:
```suggestion
    rubyVersion = rubyLine.match(/^ruby\s+(\S+)/)[1]
```
would be another way but what you did seems fine too

Issue #719, opened 2025-03-11
Unable to Parse Ruby Version from .tool-versions Since v1.222.0


Karibash:
Since version v1.222.0, the tool is unable to correctly parse the Ruby version from the `.tool-versions` file. This issue was introduced by PR #717, which changed the version parsing logic.

## Steps to Reproduce

1. Create a .tool-versions file with multiple spaces between the tool name and version:
```
flutter 3.29.1  
ruby    3.4.2  
```

2. Try to retrieve the Ruby version using the tool.

## Expected Behavior

The tool should correctly parse and return the Ruby version (3.4.2).

## Actual Behavior

The tool fails to retrieve the Ruby version when multiple spaces exist between the tool name and version.

PR #720, opened 2025-03-11
bugfix: Fix bug that version cannot be parsed in .tool-versions

Karibash:
close: #719

This PR fixes a regression introduced in PR #717, where the tool fails to correctly parse the Ruby version from .tool-versions when multiple spaces exist between the tool name and version.

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,7 +130,7 @@ function parseRubyEngineAndVersion(rubyVersion) {
   } else if (rubyVersion === '.tool-versions') { // Read from .tool-versions
     const toolVersions = fs.readFileSync('.tool-versions', 'utf8').trim()
     const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s/.test(e))[0]
-    rubyVersion = rubyLine.split(/\s/)[1]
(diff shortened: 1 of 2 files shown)


eregon (approved):
Thanks!

index.js
```diff
@@ -130,7 +130,7 @@ function parseRubyEngineAndVersion(rubyVersion) {
   } else if (rubyVersion === '.tool-versions') { // Read from .tool-versions
     const toolVersions = fs.readFileSync('.tool-versions', 'utf8').trim()
     const rubyLine = toolVersions.split(/\r?\n/).filter(e => /^ruby\s/.test(e))[0]
-    rubyVersion = rubyLine.split(/\s/)[1]
+    rubyVersion = rubyLine.split(/\s+/)[1]
```
eregon, line 133: It makes me think we should really do https://github.com/ruby/setup-ruby/pull/717/files#r1982040471 but I merged this as-sis to fix it ASAP.
We should also add a test for .tool-versions

eregon: 30755d8bf4ebfa38caa5ad3a6b808846393196a5

Issue #718, opened 2025-03-10
Choose appropriate minimum JDK for JRuby versions


headius:
JRuby 10 is just about to be released, and master (now hosting 10) will become `jruby-head` once we unlock jruby-dev-builder (https://github.com/ruby/jruby-dev-builder/pull/9). JRuby 10, unlike 9.4, depends on JDK 21 minimum. Because GHA defaults JAVA_HOME to a Java 17 install, we need to adjust that for all setup-ruby use on JRuby >= 10-snapshot.

Users of JRuby 10 should not have to perform an additional step to choose the proper JDK level, and we can't flip jruby-head builds to 10 until this is fixed.

I am not familiar enough with the codebase or GHA to know where to put this code. The basic requirement is to detect that we're setting up a JRuby version (head or 10+) that requires JDK 21, and alter the JAVA_HOME environment variable to point at the contents of the platform-appropriate `JAVA_HOME_21_*` variable (see https://github.com/actions/runner-images/discussions/9266 for some discussion on that).

eregon:
> I am not familiar enough with the codebase or GHA to know where to put this code

Please take a look and make a PR yourself, it should be fairly simple.
See https://github.com/ruby/setup-ruby/blob/277ba2a127aba66d45bad0fa2dc56f80dbfedffa/index.js#L55-L59
so the installer for JRuby is always ruby-builder.js: https://github.com/ruby/setup-ruby/blob/master/ruby-builder.js
It seems best to set JAVA_HOME just after https://github.com/ruby/setup-ruby/blob/277ba2a127aba66d45bad0fa2dc56f80dbfedffa/ruby-builder.js#L50

As said in https://github.com/ruby/setup-ruby/discussions/684#discussioncomment-11764702 the logic should be:
> IOW I think ideally we would detect if JAVA_HOME is the default, if it is switch to JAVA_HOME_21_X64 instead if available, for all other cases do nothing.

Regardless of the JRuby version if possible, to make it simpler (older JRuby should run fine on JDK 21, maybe even faster).

headius:
> Regardless of the JRuby version if possible

I am reluctant to switch all existing JRuby 9.4 users to Java 21. It will run fine, and may be faster, but they did not opt into such a change.

However, if it becomes a problem people are concerned about, we can address it then. Hopefully most will not notice.

eregon:
PR merged, thanks.

headius:
[...]

PR #721, opened 2025-03-11
Switch JAVA_HOME to 21 for JRuby

headius:
JRuby 10 requires Java 21. Since the previous default was 17 and all JRuby releases should work fine on 21, we do this for all JRuby installs.

Implements #718

### DIFF
--- a/common.js
+++ b/common.js
@@ -406,7 +406,11 @@ export function setupPath(newPathEntries) {
 
 export function setupJavaHome() {
   core.startGroup(`Modifying JAVA_HOME for JRuby`)
-  let newHomeVar = `JAVA_HOME_21_${os.arch().toUpperCase()}`;
+  let arch = os.arch();
+  if (arch == "x64") {
+    arch = "X64"
+  }
+  let newHomeVar = `JAVA_HOME_21_${arch}`;
   let newHome = process.env[newHomeVar];
   console.log(`Setting JAVA_HOME to ${newHomeVar} path ${newHome}`)
   core.exportVariable("JAVA_HOME", newHome);
--- a/common.js
+++ b/common.js
@@ -7,6 +7,7 @@ const crypto = require('crypto')
 const core = require('@actions/core')
 const tc = require('@actions/tool-cache')
 const exec = require('@actions/exec')
+const common = require('./common')
 const { performance } = require('perf_hooks')
 const linuxOSInfo = require('linux-os-info')
 
@@ -406,33 +407,31 @@ export function setupPath(newPathEntries) {
 }
 
 export async function setupJavaHome() {
-  core.startGroup(`Modifying JAVA_HOME for JRuby`)
+  common.measure("Modifying JAVA_HOME for JRuby", async () => {
+    console.log("attempting to run with existing JAVA_HOME")
 
-  console.log("attempting to run with existing JAVA_HOME")
+    let ret = await exec.exec('ruby', ['--version'])
 
-  let ret = await exec.exec('ruby', ['--version'])
+    if (ret === 0) {
+      console.log("JRuby successfully starts, using existing JAVA_HOME")
+    } else {
+      console.log("JRuby failed to start, try Java 21 envs")
 
-  if (ret === 0) {
-    console.log("JRuby successfully starts, using existing JAVA_HOME")
-  } else {
-    console.log("JRuby failed to start, try Java 21 envs")
+      let arch = os.arch()
+      if (arch === "x64" || os.platform() !== "darwin") {
+        arch = "X64"
+      }
 
-    let arch = os.arch()
-    if (arch === "x64" || os.platform() !== "darwin") {
-      arch = "X64"
+      let newHomeVar = `JAVA_HOME_21_${arch}`
+      let newHome = process.env[newHomeVar]
+
+      if (newHome === "undefined") {
+        throw new Error(`JAVA_HOME is not Java 21+ needed for JRuby and \$${newHomeVar} is not defined`)
+      }
+
+      console.log(`Setting JAVA_HOME to ${newHomeVar} path ${newHome}`)
+
+      core.exportVariable("JAVA_HOME", newHome)
     }
-
-    let newHomeVar = `JAVA_HOME_21_${arch}`
-    let newHome = process.env[newHomeVar]
-
-    if (newHome === "undefined") {
-      throw new Error(`JAVA_HOME is not Java 21+ needed for JRuby and \$${newHomeVar} is not defined`)
-    }
-
-    console.log(`Setting JAVA_HOME to ${newHomeVar} path ${newHome}`)
-
-    core.exportVariable("JAVA_HOME", newHome)
-  }
-
-  core.endGroup()
+  })
 }
--- a/common.js
+++ b/common.js
@@ -1,5 +1,3 @@
-import exec from "@actions/exec";
-
 const os = require('os')
 const path = require('path')
 const fs = require('fs')
@@ -8,6 +6,7 @@ const stream = require('stream')
 const crypto = require('crypto')
 const core = require('@actions/core')
 const tc = require('@actions/tool-cache')
+const exec = require('@actions/exec')
 const { performance } = require('perf_hooks')
 const linuxOSInfo = require('linux-os-info')
 
--- a/common.js
+++ b/common.js
@@ -407,7 +407,7 @@ export function setupPath(newPathEntries) {
 export function setupJavaHome() {
   core.startGroup(`Modifying JAVA_HOME for JRuby`)
   let arch = os.arch();
-  if (arch == "x64") {
+  if (arch == "x64" || os.platform() != "darwin") {
     arch = "X64"
   }
   let newHomeVar = `JAVA_HOME_21_${arch}`;
--- a/common.js
+++ b/common.js
@@ -1,3 +1,5 @@
+import exec from "@actions/exec";
+
 const os = require('os')
 const path = require('path')
 const fs = require('fs')
@@ -404,15 +406,29 @@ export function setupPath(newPathEntries) {
   return msys2Type
 }
 
-export function setupJavaHome() {
+export async function setupJavaHome() {
   core.startGroup(`Modifying JAVA_HOME for JRuby`)
-  let arch = os.arch();
-  if (arch == "x64" || os.platform() != "darwin") {
-    arch = "X64"
+
+  console.log("attempting to run with existing JAVA_HOME")
+  let ret = await exec.exec('ruby', ['--version']);
+
+  if (ret === 0) {
+    console.log("JRuby successfully starts, using existing JAVA_HOME")
+  } else {
+    console.log("JRuby failed to start, try Java 21 envs")
+    let arch = os.arch();
+    if (arch == "x64" || os.platform() != "darwin") {
+      arch = "X64"
+    }
+    let newHomeVar = `JAVA_HOME_21_${arch}`;
+    let newHome = process.env[newHomeVar];
+
+    if (newHome === "undefined") {
+      throw new Error(`JAVA_HOME is not Java 21+ needed for JRuby and \$${newHomeVar} is not defined`)
+    }
+    console.log(`Setting JAVA_HOME to ${newHomeVar} path ${newHome}`)
+    core.exportVariable("JAVA_HOME", newHome)
   }
-  let newHomeVar = `JAVA_HOME_21_${arch}`;
-  let newHome = process.env[newHomeVar];
-  console.log(`Setting JAVA_HOME to ${newHomeVar} path ${newHome}`)
-  core.exportVariable("JAVA_HOME", newHome);
+
   core.endGroup()
 }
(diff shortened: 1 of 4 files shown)


headius:
@eregon This is a quick attempt to implement what I need.

I am not sure how to detect if the user has already reassigned JAVA_HOME, either manually or by setup-java. It may not be possible.

I am trusting that `os.arch()` returns `x64` or `arm64` as expected elsewhere.

Not sure how to make this more robust.

headius:
FWIW GitHub may soon bump the default Java to 21, at which point we won't need this anymore.

common.js
```diff
@@ -403,3 +403,12 @@ export function setupPath(newPathEntries) {
   core.addPath(newPath.join(path.delimiter))
   return msys2Type
 }
+
+export function setupJavaHome() {
+  core.startGroup(`Modifying JAVA_HOME for JRuby`)
+  let newHomeVar = `JAVA_HOME_21_${os.arch().toUpperCase()}`;
```
eregon, line 409: According to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#java it's JAVA_HOME_21_arm64 on macOS arm64 and indeed it is:
https://github.com/eregon/actions-shell/actions/runs/13798182243/job/38594765321

And seeing https://github.com/eregon/actions-shell/actions/runs/13798182243/job/38594765318#step:4:11 it's `JAVA_HOME_21_X64` on linux-aarch64, that makes little sense but it is what it is.

eregon: I looked at the results, it's always `JAVA_HOME_21_X64` except on macOS-arm64 where it's `JAVA_HOME_21_arm64` instead.

eregon:
> FWIW GitHub may soon bump the default Java to 21, at which point we won't need this anymore.

Interesting, any issue I can follow about that?

common.js
headius: Does not appear there are any CI jobs for Linux on AArch64. I caught the bug in "arm64" vs "ARM64" and fixed it for macos. I guess it will also need to check if it's macos and only use the arm64 version there.

headius:
> any issue I can follow about that

No, I just assume they eventually move the default version up. Obviously they moved it up to 17 at some point.

headius:
Latest run appears to be green and handles arch properly on macos. Ruby version shows JRuby is running with 21.

headius:
@eregon Let me know if there's additional work needed here. Whenever this goes live, we can start unpinning jruby-head builds from 9.4.

eregon:
> I am not sure how to detect if the user has already reassigned JAVA_HOME, either manually or by setup-java. It may not be possible.

One idea given https://github.com/eregon/actions-shell/actions/runs/13798182243/job/38594765309#step:4:5
```
JAVA_HOME
/usr/lib/jvm/temurin-11-jdk-amd64
JAVA_HOME_8_X64
/usr/lib/jvm/temurin-8-jdk-amd64
JAVA_HOME_11_X64
/usr/lib/jvm/temurin-11-jdk-amd64
JAVA_HOME_21_X64
/usr/lib/jvm/temurin-21-jdk-amd64
JAVA_HOME_17_X64
/usr/lib/jvm/temurin-17-jdk-amd64
```

would be to only set it if `JAVA_HOME_21_X64` exists (e.g. might not exist on self-hosted runners) and `JAVA_HOME` is too old.
That way setup-ruby only overrides JAVA_HOME if strictly necessary.

You could get the Java version from `java -version`, from reading the `release` file (might not always be available though), or from finding out which `JAVA_HOME_$V_X64` env vars corresponds to `JAVA_HOME` and then Java version = $V.
That last one seems overkill, `java -version` seems fast enough, about 27ms when run locally.

common.js
```diff
@@ -403,3 +403,16 @@ export function setupPath(newPathEntries) {
   core.addPath(newPath.join(path.delimiter))
   return msys2Type
 }
+
+export function setupJavaHome() {
+  core.startGroup(`Modifying JAVA_HOME for JRuby`)
+  let arch = os.arch();
+  if (arch == "x64" || os.platform() != "darwin") {
```
eregon, line 410: `===`/`!==` since JS `==`/`!=` have horrible semantics and for consistency

common.js
```diff
@@ -403,3 +403,16 @@ export function setupPath(newPathEntries) {
...
+    arch = "X64"
+  }
+  let newHomeVar = `JAVA_HOME_21_${arch}`;
```
eregon, line 413: Could you remove the `;` to match the style?

ruby-builder.js
```diff
@@ -49,6 +49,10 @@ export async function install(platform, engine, version) {
   // Set the PATH now, so the MSYS2 'tar' is in Path on Windows
   common.setupPath([path.join(rubyPrefix, 'bin')])
 
+  if (engine == "jruby") {
```
eregon, line 52:
```suggestion
  if (engine === "jruby") {
```

headius:
Detecting the existing JAVA_HOME version is problematic; we'll have to parse the output of either `-version` or `-XinternalVersion` or the release file (not available on Java 8) and determine if the major version is > 21. Perhaps the better option would be to simply run `jruby -v` and if it fails we know we need to force JAVA_HOME or error out if there's no 21 env available.

eregon:
Could you add the check with `$JAVA_HOME/bin/java -version` so it's only set if the default Java is too old?
That seems a lot safer, e.g. wouldn't override a workflow setting JAVA_HOME to 25 or downgrade when e.g. some runner (self-hosted or not) has a default more recent than 21.

headius:
> Could you add the check with $JAVA_HOME/bin/java -version

I believe a more robust option is to run JRuby and only update JAVA_HOME if it fails.

eregon:
Is parsing the output of `-version` so tricky? (I don't know)
e.g.
```
$ java -version
openjdk version "21.0.6" 2025-01-21
OpenJDK Runtime Environment (Red_Hat-21.0.6.0.7-1) (build 21.0.6+7)
OpenJDK 64-Bit Server VM (Red_Hat-21.0.6.0.7-1) (build 21.0.6+7, mixed mode, sharing)
```
We could match the first line there, e.g. by matching the first number (`/\d+/`).

Using `jruby -v` there would make the logic a lot more complex and brittle so I don't think that's a good way.

eregon:
> I believe a more robust option is to run JRuby and only update JAVA_HOME if it fails.

Feel free to try it, e.g. in another PR.
I suspect it will be rather messy in practice.

headius:
Not sure why the exec call is producing that error. I am not familiar with how async/await or this exec module work and tried to mimic the other call I saw.

headius:
The new logic is working (update only if `ruby --version` fails), but of course all JRuby releases it sees are 9.4 and start up fine with Java 17. However the rest of the logic is not changed from what it did before, so it **should** still be working. I'm not sure how to test both scenarios.

eregon (commented):
This looks good, just a few more tweaks and it should be ready to be merged.

eregon:
> However the rest of the logic is not changed from what it did before, so it **should** still be working. I'm not sure how to test both scenarios.

Yes, I think there is no way to know until we get a JRuby 10 build, i.e. when we switch jruby-dev-builder to use master.
At that point we can trigger a workflow there manually and run the CI here (e.g. on a branch/draft PR) and see if everything is fine, if it's not we can revert the change in jruby-dev-builder & delete the last release there.

common.js
```diff
@@ -403,3 +404,35 @@ export function setupPath(newPathEntries) {
   core.addPath(newPath.join(path.delimiter))
   return msys2Type
 }
+
+export async function setupJavaHome() {
+  core.startGroup(`Modifying JAVA_HOME for JRuby`)
```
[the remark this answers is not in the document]
headius: d69712b

index.js
```diff
@@ -79,6 +79,10 @@ export async function setupRuby(options = {}) {
 
   const rubyPrefix = await installer.install(platform, engine, version)
 
+  if (engine === "jruby") {
+    await common.setupJavaHome()
+  }
```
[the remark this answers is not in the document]
headius: 5c0bbf7

common.js
```diff
@@ -403,3 +404,35 @@ export function setupPath(newPathEntries) {
...
+  console.log("attempting to run with existing JAVA_HOME")
+
+  let ret = await exec.exec('ruby', ['--version'])
+
+  if (ret === 0) {
+    console.log("JRuby successfully starts, using existing JAVA_HOME")
+  } else {
+    console.log("JRuby failed to start, try Java 21 envs")
+
+    let arch = os.arch();
```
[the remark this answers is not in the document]
headius: 2e82a48

headius:
Changes from latest review are in and green.

common.js
```diff
@@ -6,6 +6,8 @@ const stream = require('stream')
 const crypto = require('crypto')
 const core = require('@actions/core')
 const tc = require('@actions/tool-cache')
+const exec = require('@actions/exec')
+const common = require('./common')
```
eregon, line 10: This is a require of this file, it seems safer to avoid it, as I guess it might cause some weird stuff.
I'll fix it.

eregon (approved):

eregon:
There is some weird output e.g. in https://github.com/ruby/setup-ruby/actions/runs/13817559138/job/38654730800?pr=721#step:3:18
```
Modifying JAVA_HOME for JRuby
  attempting to run with existing JAVA_HOME
  > Print Ruby version
  /home/runner/.rubies/jruby-9.4.12.0/bin/ruby --version
  /home/runner/.rubies/jruby-9.4.12.0/bin/ruby --version
  jruby 9.4.12.0 (3.1.4) 2025-02-11 f4ab75096a OpenJDK 64-Bit Server VM 11.0.26+4 on 11.0.26+4 +jit [x86_64-linux]
  jruby 9.4.12.0 (3.1.4) 2025-02-11 f4ab75096a OpenJDK 64-Bit Server VM 11.0.26+4 on 11.0.26+4 +jit [x86_64-linux]
  JRuby successfully starts, using existing JAVA_HOME
  Took   2.78 seconds
Took   2.78 seconds
```
i.e. as if `Print Ruby version` was nested in `Modifying JAVA_HOME for JRuby`, but it shouldn't be.

The raw log gives:
```
2025-03-12T17:20:13.3657278Z ##[group]Extracting  Ruby
2025-03-12T17:20:13.3699863Z [command]/usr/bin/tar -xz -C /home/runner/.rubies -f /home/runner/work/_temp/0f9a3f7e-948e-43d0-b0dc-b3392b0fe27d
2025-03-12T17:20:13.6921000Z Took   0.33 seconds
2025-03-12T17:20:13.6924400Z ##[endgroup]
2025-03-12T17:20:13.6927040Z ##[group]Modifying JAVA_HOME for JRuby
2025-03-12T17:20:13.6929228Z attempting to run with existing JAVA_HOME
2025-03-12T17:20:13.6931648Z > Print Ruby version
2025-03-12T17:20:13.6947158Z [command]/home/runner/.rubies/jruby-9.4.12.0/bin/ruby --version
2025-03-12T17:20:13.6971579Z [command]/home/runner/.rubies/jruby-9.4.12.0/bin/ruby --version
2025-03-12T17:20:16.4676830Z jruby 9.4.12.0 (3.1.4) 2025-02-11 f4ab75096a OpenJDK 64-Bit Server VM 11.0.26+4 on 11.0.26+4 +jit [x86_64-linux]
2025-03-12T17:20:16.4678621Z jruby 9.4.12.0 (3.1.4) 2025-02-11 f4ab75096a OpenJDK 64-Bit Server VM 11.0.26+4 on 11.0.26+4 +jit [x86_64-linux]
2025-03-12T17:20:16.4771213Z JRuby successfully starts, using existing JAVA_HOME
2025-03-12T17:20:16.4771975Z Took   2.78 seconds
2025-03-12T17:20:16.4773023Z ##[endgroup]
2025-03-12T17:20:16.4773566Z Took   2.78 seconds
2025-03-12T17:20:16.4775891Z ##[group]Installing Bundler
```
So these two steps seem to interleave somehow, maybe a missing `await` then?

common.js
```diff
@@ -403,3 +405,33 @@ export function setupPath(newPathEntries) {
   core.addPath(newPath.join(path.delimiter))
   return msys2Type
 }
+
+export async function setupJavaHome() {
+  common.measure("Modifying JAVA_HOME for JRuby", async () => {
```
eregon, line 410:
```suggestion
  await measure("Modifying JAVA_HOME for JRuby", async () => {
```
Found the missing await

common.js
eregon, line 10:
```suggestion
```

eregon:
FWIW the first `ruby --version` can take a while on JRuby:
0.74s https://github.com/ruby/setup-ruby/actions/runs/13820508511/job/38664613489?pr=721
3.87s https://github.com/ruby/setup-ruby/actions/runs/13820508511/job/38664557503?pr=721
2.46s https://github.com/ruby/setup-ruby/actions/runs/13820508511/job/38664568863?pr=721

But, we do `Print Ruby version` `ruby --version` a second time after for sanity checking and clarity, and that one seems then faster: 0.33-0.37s, so it's the first `ruby --version` which is very slow.
And indeed before this change e.g. 2.20s https://github.com/ruby/setup-ruby/actions/runs/13788121949/job/38560974142

So not a big increase in overall time.

headius:
That first invocation is likely slow just because the JVM itself has to be paged in. I'll contemplate how to do this check more quickly in the future.

Thank you for the assistance.

PR #750, opened 2025-04-18
Support windows-11-arm

### LINKED ISSUE #600: Request: Support Windows on Arm64
Ruby recently added support for [windows-arm64](https://github.com/ruby/ruby/pull/8995).

The Ruby Installer team would like to support Windows on Arm64 as well, but the current ticket is blocked due to [lack of setup-ruby support for windows-arm](https://github.com/oneclick/rubyinstaller2/issues/362)

Therefore, I would like to request support. Unfortunately Github is not providing Windows ARM runners out-of-the-box. My company TableCheck would be happy to provide self-hosted Windows ARM runners until Github gives official support if it helps.

ntkme:
GitHub Actions recently added new windows-11-arm runner for OSS: https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

- Closes #600

This PR and the following adds support for windows-11-arm:

- https://github.com/ruby/setup-msys2-gcc/pull/21


ntkme:
Please note that this PR is still working in progress. It's currently at a point that installing ruby and jruby would work, but installing native extensions will fail because additional msys2 related updates are needed.

larskanis:
Thanks to notify me about the new Windows on ARM64 runners!

> `setup-ruby` does not use `devkit` from rubyinstaller so we will have to replicate a similar setup with https://github.com/ruby/setup-msys2-gcc.

This shouldn't be too difficult. Since the same MSYS2 x64 executables are used on arm64 as on x64, the `msys2.7z` file should be reusable. Only the `clangarm64` environment needs to be packaged into a 7z file equal to the `mingw64` and `ucrt64` environments. I think packaging these `clangarm64` files should even work on a x64 host. See https://www.msys2.org/wiki/arm64/

There is no need to select distinct packages regarding `gcc` on x64 and `clang` on arm64. The `gcc` packages are wrappers for the corresponding `clang` packages on arm64, so that the same package list can be installed per pacman. Only the different prefix `mingw-w64-clang-aarch64` vs. `mingw-w64-ucrt-x86_64` has to be respected. That's why the [rubyinstaller package list](https://github.com/oneclick/rubyinstaller2/blob/2270e14be46de7bb0488dd78b4200d39273a161b/lib/ruby_installer/build/components/03_dev_tools.rb#L13-L38) didn't change with the introduction of WoA.

ntkme:
> I think packaging these clangarm64 files should even work on a x64 host.

Not that easy. The way `setup-msys2-gcc` works today is that it uses the `C:/msys64` preinstalled on the x64 windows runner, and only pack the "difference" to create a minimal package for these runner. However, the windows-11-arm runner does not have `C:/msys64` preinstalled, so an "incremental" package simply won't work because all the base files are missing. We will have to provide two different packages, the existing "incremental" package for x64 and a new "complete" package for arm64. Currently testing that in my own fork.

eregon:
> However, the windows-11-arm runner does not have `C:/msys64` preinstalled

And there is no plan to add it by GitHub?

.github/workflows/test.yml
```diff
@@ -68,21 +68,35 @@ jobs:
...
+        - { os: windows-11-arm, ruby: '2.2' }
+        - { os: windows-11-arm, ruby: '2.3' }
+        - { os: windows-11-arm, ruby: '2.4' }
+        - { os: windows-11-arm, ruby: '2.5' }
+        - { os: windows-11-arm, ruby: '2.6' }
+        - { os: windows-11-arm, ruby: '2.7' }
+        - { os: windows-11-arm, ruby: '3.0' }
+        - { os: windows-11-arm, ruby: '3.1' }
+        - { os: windows-11-arm, ruby: '3.2' }
+        - { os: windows-11-arm, ruby: '3.3' }
```
eregon, line 83: These should be in their own category like `# Not available on Windows ARM` as they are a bit different than the other exclusions which are "not available on Windows any arch"

windows.js
```diff
@@ -129,7 +132,17 @@ async function installGCCTools(type, version) {
 // A subset of the MSYS2 base-devel group
 async function installMSYS2Tools() {
   const downloadPath = await common.measure(`Downloading msys2 build tools`, async () => {
-    let url = `${msys2GCCReleaseURI}/msys2-gcc-pkgs/msys2.7z`
+    let builderPlatform = null
+    if (os.arch() === 'x64') {
+      builderPlatform = 'windows-latest'
+    } else if (os.arch() === 'arm64') {
+      builderPlatform = 'windows-11-arm64'
```
eregon, line 139: With enough time probably it will be another version as the min GH-hosted runner in windows-arm64 like `windows-12-arm64` or so.
So we should name that platform `windows-arm64`

windows.js
```diff
@@ -129,7 +132,17 @@ async function installGCCTools(type, version) {
...
+    }
+    if (builderPlatform === null) {
```
eregon, line 141: Can you make it the `else` of the if/else above?

eregon (commented):
This looks great, but of course we'll need to wait that the related PRs are ready and approved as well

windows.js
ntkme: I was following the existing pattern of macos-latest vs macos-13-arm64, but I agree that since this is completely new, we can use a different pattern. Let me update this.

eregon:
It seems worth asking (via issue/discussion) or searching issues, that list includes things that probably don't exist on windows-arm64, but MSYS2 does work there so not clear why it's not included to me.

windows.js
eregon: Right, but in hindsight that was a mistake :sweat_smile: https://github.com/ruby/ruby-builder?tab=readme-ov-file#naming

ntkme:
> It seems worth asking (via issue/discussion) or searching issues, that list includes things that probably don't exist on windows-arm64, but MSYS2 does work there so not clear why it's not included to me.

https://github.com/actions/partner-runner-images/issues/88

Someone already asked, and it looks like they don't like the fact that msys2 for windows arm64 has hybird x64 emulation and arm64 native...

ntkme:
I added a commit in this PR to address the circular dependency on this PR and https://github.com/ruby/setup-msys2-gcc/pull/21.

The plan of action is:

1. Merge this PR as is. Lots of tests will fail temporarily.
2. Merge https://github.com/ruby/setup-msys2-gcc/pull/21.
3. Revert the commit `Temporarily disable installGCCTools on windows arm64 to solve chicken-and-egg issue`. Test will pass.

FYI @MSP-Greg

.github/workflows/test.yml
```diff
@@ -64,25 +64,43 @@ jobs:
         - { os: ubuntu-24.04-arm, ruby: '2.0' }
         - { os: ubuntu-24.04-arm, ruby: '2.1' }
         - { os: ubuntu-24.04-arm, ruby: '2.2' }
+        # These older Rubies do not exist on Windows arm64
```
eregon, line 67: Actually it's because RubyInstaller doesn't have builds, right? (and of course these older versions might not compile on Windows arm64).
If so, maybe `# RubyInstaller windows-arm64 builds only exist for Ruby 3.4+`

ntkme: From ruby installer’s release note:

> The ARM version required some changes to the side-by-side DLL loading mechanism and it is solved in such a way, that also fixes [a longstanding issue on x64/x86](https://github.com/oneclick/rubyinstaller2/issues/60). Unfortunately these changes are too heavy to be backported to ruby-3.3 and older, so that no ARM package of these older ruby versions will be released.

ntkme: Updated.

MSP-Greg:
@ntkme

Today I added PR's to my forks of setup-ruby and setup-msys2-gcc, and generated the two 7za files needed for arm64.

So, they now exist in the [ruby/setup-msys2-gcc release](https://github.com/ruby/setup-msys2-gcc/releases/tag/msys2-gcc-pkgs).

You should be able to remove the 'Temporarily disable installGCCTools' commit and CI should pass, as it did in my fork.

After this is accepted, we can switch the 'setup-msys2-gcc' PR back to using `ruby/setup-ruby`.

Thanks again,

Greg

ntkme:
> You should be able to remove the 'Temporarily disable installGCCTools' commit and CI should pass, as it did in my fork.

Thanks, this greatly simplifies the operation here. 

> After this is accepted, we can switch the 'setup-msys2-gcc' PR back to using ruby/setup-ruby.

That PR is already using `ruby/setup-ruby` as I found out that I can do the workaround in this PR to disable the gcc setup... But now you have uploaded the achieve needed, no workaround is needed anymore, except for that this PR to be merged before the 'setup-msys2-gcc' PR.

eregon (approved):
Thank you for all your work on this!

PR #761, opened 2025-04-30
Handle win2019 gcc issue

MSP-Greg:
Fix for GCC 15 being preinstalled on win2019

### DIFF
--- a/windows.js
+++ b/windows.js
@@ -89,7 +89,7 @@ export async function install(platform, engine, version) {
 
   // windows 2016 and 2019 need ucrt64 installed, 2022 and future images need
   // ucrt64 or mingw64 installed, depending on Ruby version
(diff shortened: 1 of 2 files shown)


eregon:
Link: https://github.com/ruby/setup-ruby/pull/760#issuecomment-2843034045

windows.js
```diff
@@ -89,7 +89,7 @@ export async function install(platform, engine, version) {
 
   // windows 2016 and 2019 need ucrt64 installed, 2022 and future images need
   // ucrt64 or mingw64 installed, depending on Ruby version
-  if (((msys2Type === 'ucrt64') || !hasMSYS2PreInstalled) && common.floatVersion(version) >= 2.4) {
+  if ((msys2Type === 'ucrt64') || (common.floatVersion(version) >= 2.4)) {
```
eregon, line 92: So this installs GCC even if MSYS2 is pre-installed, to avoid the too new gcc from the pre-installed MSYS2, right?

eregon (approved):
Thanks!

PR #762, opened 2025-05-04
Rewrite ruby/setup-msys2-gcc

ntkme:
This PR include a few changes for windows:

- Fix the issue with gcc 15/14 and mingw-w64 r688/r720 by introducing flexible variants of toolchains.
- Reduce friction with https://github.com/ruby/setup-msys2-gcc/pull/26.
- Improves performance by install to fast disk and create junction link on slow disk.
    - Ruby is installed to `$RUNNER_TEMP\$base` and a junction link is created at original `$rubyPrefix` if `$rubyPrefix` is on the slow disk.
    - MSYS2 is installed to `$RUNNER_TEMP\msys64` and a junction link is created at `$rubyPrefix\msys64`.
- Improves the reliability of `ridk enable` step by using proper serialization.
- Refactor and clean up tech debt.

### Performance

Run to run improvement varies a lot due to different sizes of packages and the unstable performance of the slow disk.
However, in general this saves anywhere from 20s to 60s per run. Example of a performance improvement from 63.43s to 27.15s - or 36.38s in saving:

https://github.com/ruby/setup-ruby/actions/runs/14772696897/job/41475457782

```
Downloading Ruby
  https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-3.4.3-1/rubyinstaller-3.4.3-1-x64.7z
  Took   0.97 seconds
Extracting  Ruby
  Took  25.38 seconds
Modifying Path
Downloading msys2 build tools
  https://github.com/ruby/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/msys2.7z
  Took   0.73 seconds
Extracting  msys2 build tools
  Took  12.52 seconds
Downloading ucrt64-3.0 build tools
  https://github.com/ruby/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/ucrt64-3.0.7z
  Took   1.00 seconds
Extracting  ucrt64-3.0 build tools
  Took  22.83 seconds
```

https://github.com/ntkme/setup-ruby/actions/runs/14823150160/job/41612838924

```
Downloading Ruby
  https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-3.4.3-1/rubyinstaller-3.4.3-1-x64.7z
  Took   0.46 seconds
Extracting  Ruby
  Took   9.87 seconds
Modifying Path
Downloading msys2 build tools
  https://github.com/ntkme/setup-msys2-gcc/releases/latest/download/msys2-ucrt64-gcc@14.7z
  Took   1.18 seconds
Extracting  msys2 build tools
  Took  15.64 seconds
```

---

Note regarding why use `$rubyPrefix\msys64`:
- `ENV["MSYS2_PATH"]` have highest priority, but it's only supported on ruby >=3.0.7: https://github.com/oneclick/rubyinstaller2/commit/a4590dd112705a9b1042d5d03fb253c8b6a5b669 
- `$rubyPrefix\msys64` has the second highest priority, and it's supported by ruby >=2.4.1: https://github.com/oneclick/rubyinstaller2/commit/21a416ffb73c876a344f08f2d362b5a381984b12

`>=2.4.1` means `$rubyPrefix\msys64` is supported by all currently available ruby installer versions that's linked with msys2, and it has higher priority than `C:\msys64` so that we don’t need to touch the preinstalled `C:\msys64` at all.


MSP-Greg:
Curious, why the need for 'Revert back to use `GITHUB_WORKSPACE` to detect fast drive'?

MSP-Greg:
@eregon

I've quickly looked as this, and everything seems ok.  We're getting close to finalizing the changes to `setup-msys2-gcc`, so checking is this is okay.  Once the new packages are created, this can be merged.

eregon:
From the PR title I thought this was a ruby/setup-msys2-gcc PR :sweat_smile: 
I'll try to take a look soon.

ntkme:
clangarm64 ruby head test is now passing after I disabled the "update bundler" step in setup-ruby. The problem is that in this repo we're intentionally setting up msys2 after installing ruby during our test. Not having msys2 installed currently causes the built-in `date` gem to crash, and the "update bundler" step loads the `date` gem.

See https://github.com/oneclick/rubyinstaller2-packages/issues/25#issuecomment-2881729508 and https://github.com/ruby/date/issues/126#issuecomment-2881751295

generate-windows-versions.rb
```diff
@@ -31,18 +31,62 @@
...
+windows_toolchain = JSON.parse(URI.open(base_url) { |f| f.read }, symbolize_names: true)
+
+versions.each do |raw_version, archs|
+  version = Gem::Version.correct?(raw_version) ? Gem::Version.new(raw_version) : head_version
+
+  archs.each_key do |raw_arch|
+    arch = raw_arch == 'arm64' ? 'aarch64' : raw_arch
+
+    platform = case raw_version
+               when 'head', 'ucrt'
```
eregon, line 63:
```suggestion
    when 'head', 'ucrt'
```
Could you align as if there was no assignment (IMO assignment should never cause indentation), or if your editor gets in the way then a single indent after `=`:
```
platform =
  case
  when
```
?

generate-windows-versions.rb
```diff
@@ -31,18 +31,62 @@
...
+               else
+                 if version >= Gem::Version.new('3.1.0.dev')
+                   "#{arch}-mingw-ucrt"
+                 else
+                   "#{arch}-mingw32"
+                 end
+               end
+
+    toolchain = windows_toolchain
+      .select do |pkg|
```
eregon, line 78:
```suggestion
      windows_toolchain.select do |pkg|
```
(and no windows_toolchain on previous line)

windows-toolchain-versions.json
```diff
@@ -0,0 +1,258 @@
...
+    "x64": "https://github.com/oneclick/rubyinstaller/releases/download/devkit-4.7.2/DevKit-mingw64-64-4.7.2-20130224-1432-sfx.exe"
+  },
+  "2.3.1": {
+    "x64": "https://github.com/oneclick/rubyinstaller/releases/download/devkit-4.7.2/DevKit-mingw64-64-4.7.2-20130224-1432-sfx.exe"
+  },
+  "2.3.3": {
+    "x64": "https://github.com/oneclick/rubyinstaller/releases/download/devkit-4.7.2/DevKit-mingw64-64-4.7.2-20130224-1432-sfx.exe"
+  },
+  "2.4.1": {
+    "x64": "https://github.com/ntkme/setup-msys2-gcc/releases/latest/download/msys2-mingw64-mingw-w64@r688-gcc@14-openssl@1.1.7z"
```
eregon, line 21:
```suggestion
    "x64": "https://github.com/ruby/setup-msys2-gcc/releases/latest/download/msys2-mingw64-mingw-w64@r688-gcc@14-openssl@1.1.7z"
```
after that PR is merged

windows.js
```diff
@@ -65,230 +37,191 @@ export async function install(platform, engine, version) {
...
-  await common.measure('Extracting  Ruby', async () =>
-    // -xr extract but exclude share\doc files
-    exec.exec('7z', ['x', downloadPath, '-bd', `-xr!${base}\\share\\doc`, `-o${parentDir}`], { silent: true }))
+  const extractPath = path.join(process.env.RUNNER_TEMP, 'msys64')
+  await common.measure('Extracting  msys2 build tools', async () =>
+    exec.exec('7z', ['x', downloadPath, '-aoa', '-bd', `-o${extractPath}`], { silent: true }))
 
-  if (base !== path.basename(rubyPrefix)) {
-    await io.mv(path.join(parentDir, base), rubyPrefix)
+  const msys2Path = path.join(rubyPrefix, 'msys64')
```
eregon, line 123: Could you add a code comment here about the choice of rubyPrefix/msys64, like you explained in the PR description?

eregon (approved):
This looks great, thank you for your work on this!

I guess the order is merge https://github.com/ruby/setup-msys2-gcc/pull/26, create the release & assets, then update this PR to use ruby/setup-msys2-gcc links, and merge this PR?

ntkme:
> I guess the order is merge https://github.com/ruby/setup-msys2-gcc/pull/26, create the release & assets, then update this PR to use ruby/setup-msys2-gcc links, and merge this PR?

Correct.

MSP-Greg:
@eregon

> I guess the order is merge https://github.com/ruby/setup-msys2-gcc/pull/26, create the release & assets, then update this PR to use ruby/setup-msys2-gcc links, and merge this PR?

The PR has been merged in https://github.com/ruby/setup-msys2-gcc

ntkme:
@eregon Release artifacts has been uploaded and I've updated this PR to use ruby/setup-msys2-gcc. There is a truffleruby test that's failing but it's unrelated to my change.

ronaldtse:
Hoping this gets merged soon as we have a number of release pipelines blocked due to the recent Ruby 3.1 on Windows issue. Thank you for the excellent work and maintenance! 🙇‍♂️

eregon:
A bit more debugging in https://github.com/eregon/actions-shell/actions/runs/15309439156/job/43070305903#step:4:550 shows `D:/a/_temp/` paths in $" but `C:/hostedtoolcache/windows/Ruby/3.4.4/x64/` in $:, which seems not unlikely to cause some incompatibility in gem test suites.
And indeed `File.realpath("C:/hostedtoolcache/windows/Ruby/3.4.4/x64/lib/ruby/site_ruby/3.4.0") = "D:/a/_temp/rubyinstaller-3.4.4-2-x64/lib/ruby/site_ruby/3.4.0"`

ntkme:
> which seems not unlikely to cause some incompatibility in gem test suites.

Yes, I understand the junction can cause issues when some code use a mix of `realpath` vs `absolute_path`. The main reason to use junction is to let us extract files on the faster disk.

I wish the code run into the issue can be more consistent in that either to use `realpath` everywhere or use `absolute_path` everywhere but never a mix of the two, but not sure if that's a fair ask. If we want to pursue better compatibility than performance, we can easily revert the usage of junction and just extract to slow disk as before.

MSP-Greg:
@eregon

IIRC, Ruby head builds have always used D:\ for their `PATH` addition, so it should work fine.  Rubies that are part of the runner image will need to stay in `C:/hostedtoolcache/windows/Ruby`.

I probably can't get to it tonite, tomorrow should be better...

Lastly, I also recall that there are two kinds of Windows links, may look at that also.

ntkme:
> Lastly, I also recall that there are two kinds of Windows links, may look at that also.

Correct, there are two kinds of soft links: junction and symlink. There is also hard link but hard link can not have directory as target.

There is a good explanation about the difference here: https://superuser.com/questions/343074/directory-junction-vs-directory-symbolic-link.

The biggest concern I had at the time was that creating symlink require privileged access, while junction only need regular access.

ntkme:
Another idea here:

Detect if we're in self-hosted runner - If so extract into hosttoolcache (no junction) directly, otherwise extract into `$RUNNER_TEMP` (also no junction) directly.

eregon:
> Detect if we're in self-hosted runner - If so extract into hosttoolcache (no junction) directly, otherwise extract into `$RUNNER_TEMP` (also no junction) directly.

It should work, but since no one complained and probably no one uses self-hosted runners on Windows with persistent toolcache let's keep it simple and do only that latter part always.
FYI detecting self-hosted runner is unfortunately not so reliable unfortunately.

ntkme:
> FYI detecting self-hosted runner is unfortunately not so reliable unfortunately.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

> RUNNER_ENVIRONMENT

> The environment of the runner executing the job. Possible values are: `github-hosted` for GitHub-hosted runners provided by GitHub, and self-hosted for `self-hosted` runners configured by the repository owner.

eregon:
Interesting, but in reality it's more complicated than that as there are some self-hosted runners for which we can reuse prebuilt Rubies and some self-hosted runners where we can't. Anyway, simple is almost always better unless there is a need for something more complicated (I don't see one).

PR #767, opened 2025-05-18
Upgrade @actions/cache to 4.0.3

JoeDupuis:
Upgrade @actions/cache to fix `Failed to save: Failed to CreateCacheEntry: Received non-retryable error` errors.

Screenshot 2025-05-18 at 1 57 55 AM

Here's a test run with a rails app before the fix: https://github.com/JoeDupuis/test-setup-ruby-fix/actions/runs/15094257150

and after the fix: https://github.com/JoeDupuis/test-setup-ruby-fix/actions/runs/15094275803

The error only show when trying to bust the cache.

More context on [this actions/cache issue](https://github.com/actions/cache/issues/1541)


eregon:
From https://github.com/actions/cache/issues/1541#issuecomment-2729883487 it seems to mention 4.2.2, ah but that's actions/cache, not the toolkit version with the fix, which seems to be 4.0.2 from https://github.com/actions/cache/releases/tag/v4.2.2.
And https://github.com/actions/toolkit/blob/main/packages/cache/package.json#L3 is 4.0.3.
OK, thank you.

BTW it's a warning not an error but yes good to avoid unfixable warnings.

PR #769, opened 2025-05-20
Update/upgrade packages

MSP-Greg:
In the current [`yarn.lock`](https://github.com/ruby/setup-ruby/blob/c95ae3725f6ebdd095f2bd19caed7ebc14435ba5/yarn.lock) file, all packages are from `registry.npmjs.org`.

In the previous commit's [file](https://github.com/ruby/setup-ruby/blob/cb0fda56a307b8c78d38320cd40d9eb22a3bf04e/yarn.lock), all packages are from `registry.yarnpkg.com`.

Not sure if this was intentional.

Commits:
1. 'package.json - update @actions/io, semver, ncc' - updated ncc and other packages to show currently used packages.
2. 'yarn.lock update' - run the command and install
3. 'run yarn `run package`' - updated dist/index.js file.  It grows by about 3%.


eregon:
> In the current [`yarn.lock`](https://github.com/ruby/setup-ruby/blob/c95ae3725f6ebdd095f2bd19caed7ebc14435ba5/yarn.lock) file, all packages are from `registry.npmjs.org`.
> 
> In the previous commit's [file](https://github.com/ruby/setup-ruby/blob/cb0fda56a307b8c78d38320cd40d9eb22a3bf04e/yarn.lock), all packages are from `registry.yarnpkg.com`.
> 
> Not sure if this was intentional.

Ah right it's from https://github.com/ruby/setup-ruby/pull/767 I didn't notice.
I guess it doesn't matter much either way.
Not sure how it can happen though.

MSP-Greg:
@eregon

> I guess it doesn't matter much either way.

Yes, I thought it would make future commits that were bug/feature changes a bit less noisy.  Also, some things needed updates.

I found some old 'articles' that stated that `registry.yarnpkg.com` just redirected to `registry.npmjs.org`.  A quick check with `curl` seemed to show that is no longer the case...

PR #780, opened 2025-07-16
Add workaround for Ruby 3.1 after RubyGems 3.7.0 release

deivid-rodriguez:
Like every time I release a version of RubyGems dropping support for a EOL Ruby, I forgot to update `ruby/setup-ruby` before the release, so that `rubygems: latest` does not break 😅

Fixing it now.

This issue was fixed by https://github.com/rubygems/rubygems/pull/7334, shipped with RubyGems 3.5.4. That version of RubyGems comes by default with Ruby 3.3.1 or higher, so I believe we just need one more future update similar to this one right before shipping a RubyGems version dropping support for Ruby 3.2. I will try to remember that one last time!

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -137,9 +137,8 @@ export async function installBundler(bundlerVersionInput, rubygemsInputSet, lock
   }
 
   const gem = path.join(rubyPrefix, 'bin', 'gem')
-  // Workaround for https://github.com/rubygems/rubygems/issues/5245
   // and for https://github.com/oracle/truffleruby/issues/2780
-  const force = ((platform.startsWith('windows-') && engine === 'ruby' && floatVersion >= 3.1) || (engine === 'truffleruby')) ? ['--force'] : []
+  const force = engine === 'truffleruby' ? ['--force'] : []
 
   const versionParts = [...bundlerVersion.matchAll(/\d+/g)].length
   const bundlerVersionConstraint = versionParts >= 3 ? bundlerVersion : `~> ${bundlerVersion}.0`
--- a/bundler.js
+++ b/bundler.js
@@ -137,13 +137,11 @@ export async function installBundler(bundlerVersionInput, rubygemsInputSet, lock
   }
 
   const gem = path.join(rubyPrefix, 'bin', 'gem')
-  // and for https://github.com/oracle/truffleruby/issues/2780
-  const force = engine === 'truffleruby' ? ['--force'] : []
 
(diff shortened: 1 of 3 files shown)


simi (commented):
:muscle:

deivid-rodriguez:
Uggh, apparently on Windows Bundler is installed with `--force` flag to `gem install`, and that bypasses any dependency constraints (including `required_ruby_version`). So the new release has also broken something on Windows. Looking into it.

deivid-rodriguez:
Not using the `--force` flag on Windows fixes things. It was added apparently to fix some CI issues in https://github.com/ruby/setup-ruby/pull/253, but nothing is failing now without it so I assume real culprit was fixed somewhere (maybe rubyinstaller2).

bundler.js
```diff
@@ -137,9 +137,8 @@ export async function installBundler(bundlerVersionInput, rubygemsInputSet, lock
   }
 
   const gem = path.join(rubyPrefix, 'bin', 'gem')
-  // Workaround for https://github.com/rubygems/rubygems/issues/5245
   // and for https://github.com/oracle/truffleruby/issues/2780
-  const force = ((platform.startsWith('windows-') && engine === 'ruby' && floatVersion >= 3.1) || (engine === 'truffleruby')) ? ['--force'] : []
+  const force = engine === 'truffleruby' ? ['--force'] : []
```
eregon, line 141: Could you try removing this line and the comment entirely?
I think at the source it was a RubyGems bug to be too careful about conflicts with `bin/bundle`.
And I guess probably the fix got backported to the recent 3.1 releases.

eregon (commented):
Thank you!

bundler.js
eregon: > I think at the source it was a RubyGems bug to be too careful about conflicts with `bin/bundle`.
> And I guess probably the fix got backported to the recent 3.1 releases.

Correction: reading https://github.com/oracle/truffleruby/issues/2780#issuecomment-1308981564 more carefully, the problem happened when updating to latest RubyGems, which added a regression with https://github.com/rubygems/rubygems/pull/2929/files#diff-b25097522a961ed7158012499383b8a8057caf87cbc445d6b168a5b8a27ba45dR224 and that got fixed in newer RubyGems versions.
So we shouldn't need this workaround at all anymore, because anyway it only happened when using latest rubygems and latest rubygems no longer has this issue (I assume).

deivid-rodriguez: Done! It should be good now if CI is fine.

eregon (approved):

PR #785, opened 2025-07-24
Add printing of the lockfile after installation

### LINKED ISSUE #783: Print the final content of the Gemfile.lock
Feature request: Print the final content of the Gemfile.lock.

From a diagnostics point of view, it is very useful to see exactly what versions of gems were installed from run to run.

In our particular case, we had a corrupted cache which caused the list of gems to be rebuilt, and the new run failed our CI. While I could see the newly installed gems as they were installed, when I went to the previous green run, I couldn't see what was installed.

The request is simply, at the end of installation, print the Gemfile.lock into the log in a collapsed section.

I am willing to make a PR for this feature but wanted to discuss first.

Fryguy:
Closes #783

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,9 @@ export async function setupRuby(options = {}) {
   if (inputs['bundler-cache'] === 'true') {
     await common.time('bundle install', async () =>
       bundler.bundleInstall(gemfile, lockFile, platform, engine, version, bundlerVersion, inputs['cache-version']))
+
+    await core.group(`Print lockfile`, async () =>
+      await exec.exec('cat', [lockFile]))
   }
 
   core.setOutput('ruby-prefix', rubyPrefix)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -99,6 +99,9 @@ export async function setupRuby(options = {}) {
   if (inputs['bundler-cache'] === 'true') {
     await common.time('bundle install', async () =>
       bundler.bundleInstall(gemfile, lockFile, platform, engine, version, bundlerVersion, inputs['cache-version']))
+
+    await common.measure(`Print ${lockFile}`, async () =>
```
eregon, line 103:
```suggestion
    await core.group`Print ${lockFile}`, async () =>
```
So there is no `Took   0.01 seconds`

eregon (approved):
Thanks!

eregon:
@Fryguy See https://github.com/ruby/setup-ruby/commit/e8621f0b86851d7377d85882bb751b24dfadd958#commitcomment-162888500
Could you make the PR again and address that, and also adding a test for that case of using `bundler-cache: true` but having no Gemfile?

Fryguy:
Oh interesting - I never expected lockfile to be null because it is passed to bundler.bundleInstall immediate before it.  I'll add a test case for that.

Issue #788, opened 2025-07-25
Can no longer run without a lockfile


taustgen:
### Ensure the following before filing this issue

- [x] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [x] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

### Are you running on a GitHub-hosted runner or a self-hosted runner?

GitHub-hosted runner

### Link to the failed workflow job (must be a public workflow job, so the necessary information is available)

Close this one if it really matters

### Any other notes?

https://github.com/ruby/setup-ruby/commit/e8621f0b86851d7377d85882bb751b24dfadd958#commitcomment-162887376

The above change needs to be wrapped in a conditional to make sure `lockfile` exists

eregon:
I reverted that change for now in https://github.com/ruby/setup-ruby/pull/789.

eregon:
So I think that means you use `bundler-cache: true` but you don't have a Gemfile?
That's a bit of a weird case, but since it worked before yes it makes sense to not break it: https://github.com/ruby/setup-ruby/blob/0ecad18fe538ef70f6b82773daecc6af1a7fe58a/bundler.js#L151

Issue #787, opened 2025-07-25
Please make #783 Optional


aservedio-swi:
### Ensure the following before filing this issue

- [x] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [x] I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)

### Are you running on a GitHub-hosted runner or a self-hosted runner?

self-hosted runner

### Link to the failed workflow job (must be a public workflow job, so the necessary information is available)

Will copy in step that failed

### Any other notes?

For our self-hosted runners, #783 is failing as it is not finding (most likely does not have permission to run) `cat`.

Can this feature be made optional?

The action was configured as:
```yaml
      - name: Set up Ruby 2.7
        uses: ruby/setup-ruby@v1
        with:
          # Not needed with a .ruby-version file
          ruby-version: 2.7.6
          # The version of RubyGems to use. Either 'default' (the default), 'latest', or a version number (e.g., 3.3.5).
          rubygems: 3.2.3
          # runs 'bundle install' and caches installed gems automatically
          bundler-cache: true
```

Output:
```
Run ruby/setup-ruby@v1
  with:
    ruby-version: 2.7.6
    rubygems: 3.2.3
    bundler-cache: true
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.13/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.13/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.13/x64/lib
Modifying PATH
  Entries added to PATH to use selected Ruby:
    /opt/hostedtoolcache/Ruby/2.7.6/x64/bin
Downloading Ruby
  https://github.com/ruby/ruby-builder/releases/download/toolcache/ruby-2.7.6-ubuntu-22.04.tar.gz
[...]

PR #790, opened 2025-07-25
Add printing of the lockfile after installation (retry)

### LINKED ISSUE #783: Print the final content of the Gemfile.lock

Fryguy:
Reimplementation of #785, but with checking the case of a missing lockfile (caused by a missing Gemfile).

While the case of user `bundler-cache: true` with a missing Gemfile is strange, it is something that was working prior to #785.

Closes #783

@eregon Please review.

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,11 @@ export async function setupRuby(options = {}) {
   if (inputs['bundler-cache'] === 'true') {
     await common.time('bundle install', async () =>
       bundler.bundleInstall(gemfile, lockFile, platform, engine, version, bundlerVersion, inputs['cache-version']))
+
+    if (fs.existsSync(lockFile)) {
+      await core.group(`Print lockfile`, async () =>
+        await exec.exec('cat', [lockFile]))
+    }
   }
 
   core.setOutput('ruby-prefix', rubyPrefix)
--- a/index.js
+++ b/index.js
@@ -100,7 +100,7 @@ export async function setupRuby(options = {}) {
     await common.time('bundle install', async () =>
       bundler.bundleInstall(gemfile, lockFile, platform, engine, version, bundlerVersion, inputs['cache-version']))
 
-    if (fs.existsSync(lockFile)) {
+    if (lockFile !== null && fs.existsSync(lockFile)) {
       await core.group(`Print lockfile`, async () =>
         await exec.exec('cat', [lockFile]))
     }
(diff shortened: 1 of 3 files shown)


Fryguy:
I verified with the new test case (via nektos/act) that before the code change it exhibited the failures that users were seeing in #787 and #788, and after the changes to check the lockfile exists, that case works: https://github.com/ruby/setup-ruby/actions/runs/16531408828/job/46757430272?pr=790#step:4:32

Before:

```
| > bundle install
| Could not determine gemfile path, skipping "bundle install" and caching
| Took   0.00 seconds
[Test this action/Test with no Gemfile but with bundler-cache]   ❓  ::group::Print lockfile
| [command]/usr/bin/cat null
| /usr/bin/cat: null: No such file or directory
[Test this action/Test with no Gemfile but with bundler-cache]   ❓  ::endgroup::
[Test this action/Test with no Gemfile but with bundler-cache]   ❗  ::error::The process '/usr/bin/cat' failed with exit code 1
[Test this action/Test with no Gemfile but with bundler-cache]   ❌  Failure - Main ./
```

After

```
| > bundle install
| Could not determine gemfile path, skipping "bundle install" and caching
| Took   0.00 seconds
[Test this action/Test with no Gemfile but with bundler-cache]   ✅  Success - Main ./
```

index.js
```diff
@@ -99,6 +99,11 @@ export async function setupRuby(options = {}) {
   if (inputs['bundler-cache'] === 'true') {
     await common.time('bundle install', async () =>
       bundler.bundleInstall(gemfile, lockFile, platform, engine, version, bundlerVersion, inputs['cache-version']))
+
+    if (fs.existsSync(lockFile)) {
```
eregon, line 103:
```suggestion
    if (lockFile !== null && fs.existsSync(lockFile)) {
```
It feels a bit weird to call `fs.existsSync(null)` so let's be more explicit

eregon (approved):
Thank you!

PR #807, opened 2025-09-20
Reinstate testing against jruby 9.4 and head

chadlwilson:
- Reverts https://github.com/ruby/setup-ruby/commit/f2f42b7848feff522ffa488a5236ba0a73bccbdd now the issue is fixed upstream
- Reinstates testing against JRuby 9.4 (I believe when this was last reviewed 10.0 was unreleased so `jruby` was testing 9.4 and `jruby-head` was testing 10.0)

If you're not comfortable adding yet more jobs, can reconsider adding 9.4. Having said that, given the significant changes and major Java version increase, possibly wise to test both given they are both supported.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -195,8 +195,7 @@ jobs:
       shell: bash
       run: echo ~ && bundle install
     - name: Windows JRuby
-      # Should be startsWith(matrix.ruby, 'jruby') but broken on jruby-head: https://github.com/jruby/jruby/issues/8623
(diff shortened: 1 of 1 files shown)


eregon:
> Having said that, given the significant changes and major Java version increase, possibly wise to test both given they are both supported.

In general the de-facto rule so far was to only test latest release for non-CRuby, but this seems a good enough argument, let's do it, at least for now and while it doesn't cause issues.

eregon (approved):
Thank you for the PR

PR #812, opened 2025-09-30
Skip saving cache for merge_group event

### LINKED ISSUE #805: Provide ability to restore from cache but NOT save cache
Hello 👋 

I am hoping to discuss whether you'd consider adding inputs that would allow a caller to indicate that they'd like to _restore_ bundler cache but NOT save it at the end of the workflow.

**Rationale:** In _some_ scenarios, it does not make sense to save the cache. One such scenario is when PRs are running in the merge queue. The branch associated with the PRs being checked looks like `gh-readonly-queue/main/pr-15190-f5aefc512c3719594eb22d601d2824ad519d07a0`. This is scoped to a PR _and_ a commit SHA. Due to the way cache restoration via `actions/cache` works, this cache will only be used by subsequent runs on the same PR _and_ the same sha _only_ in the queue (which is extremely unlikely, if not impossible--I'm not sure if that SHA is the result of several PRs being combined or what). Anyway, I'd love to avoid adding a bunch of these useless cached objects to the cache and increasing the likelihood of cache thrashing, if at all possible.

**Request:** I am wondering if it would be as simple as adding a parameter `save-cache: true|false` to the action:

```yaml
- name: Setup Ruby and Bundler
  if: ${{ inputs.install_ruby == 'true'}}
  uses: ruby/setup-ruby@44511735964dcb71245e7e55f72539531f7bc0eb # v1.257.0
  with:
    ruby-version: ${{ steps.grab-versions.outputs.ruby_version }}
    bundler-cache: true
    working-directory: ${{ inputs.bundler_working_directory }}
    save-cache: ${{ github.event_name != "merge_group" }}
```
_For ultimate clarity, parameters `restore-bundler-cache` and `save-bundler-cache` would be better._

This is akin to the approach you can take when using `actions/cache/restore` and `actions/cache/save` directly. We do this in other places within our workflows.

ntkme:
- Closes #805

Note, I did not do `yarn update` in order to keep the lockfile change small. This has introduced some transitive dependencies at different versions. One of the maintainers should probably do a dependency update at some point.

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -218,7 +218,8 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
   await exec.exec('bundle', ['install', '--jobs', `${jobs}`])
 
   // @actions/cache only allows to save for non-existing keys
-  if (!common.isExactCacheKeyMatch(key, cachedKey)) {
+  // Also, skip saving cache for merge_group event
+  if (!common.isExactCacheKeyMatch(key, cachedKey) && process.env['GITHUB_EVENT_NAME'] !== 'merge_group') {
     if (cachedKey) { // existing cache but Gemfile.lock differs, clean old gems
       await exec.exec('bundle', ['clean'])
     }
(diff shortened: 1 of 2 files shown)


ntkme:
@eregon Would you mind take a look and see if this makes sense?

In #805 we have a user complaining that `merge_group` (merge queue) events are saving cache, but those cache will never have any chance to be read again. In a large repository, this can consume lots of cache space on GitHub to the point it is purging older cache aggressively, and significantly reduce overall cache hit rate.

PR #831, opened 2025-11-26
add option to set config BUNDLE_FROZEN for bundler install

KyFaSt:
* Adds support for a new bundle-frozen option in the Ruby setup action whic enforces that Gemfile.lock remains unchanged during bundle install for stricter dependency consistency.
* Tried to be flexible to allow implementation of other config options in future

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -437,14 +437,6 @@ jobs:
         gem 'rake', '~> 13.0'
         gem 'minitest', '~> 5.16'
         EOF
-    - name: Generate Gemfile.lock
-      shell: bash
-      run: |
-        gem install bundler
-        bundle lock
-    - name: Display Gemfile.lock
-      shell: bash
-      run: cat Gemfile.lock
     - uses: ./
       with:
         ruby-version: ${{ matrix.ruby }}


eregon:
Isn't this already the case with https://github.com/ruby/setup-ruby/blob/8aeb6ff8030dd539317f8e1769a044873b56ea71/bundler.js#L162 ?
Maybe something that changed in a recent Bundler version?

mrecachinas:
@eregon Yes, that does achieve the same goal. I think the issue is if no lockfile is provided, it will currently hit the [`else` branch](https://github.com/ruby/setup-ruby/blob/8aeb6ff8030dd539317f8e1769a044873b56ea71/bundler.js#L164-L166) and generate one. It seems the intent here is to force `BUNDLE_FROZEN=true` as a config option on the Action so it will fail if there's no lockfile (or the lockfile differs from the Gemfile).

eregon:
AFAIK `BUNDLE_FROZEN=true` means nothing if there is no lockfile already.

Closing this, as there seems to be no use case that needs it, and frozen is already always the case when there is a lockfile.

eregon:
Looking at the added test it's for a second `bundle install` done later, but in that case it's much simpler to just `bundle config set frozen true` yourself after setup-ruby.

mrecachinas:
> AFAIK BUNDLE_FROZEN=true means nothing if there is no lockfile already.

It'll error if there's no associated lockfile and you set `BUNDLE_FROZEN=true` or equivalent (`bundle config set frozen true`).

```bash
~/Projects/test-rb  v3.4.2
$ ls -latr
total 8
drwxr-xr-x  28 mrecachinas  staff  896 Jan  4 08:47 ..
-rw-r--r--   1 mrecachinas  staff   14 Jan  4 08:48 Gemfile
drwxr-xr-x   3 mrecachinas  staff   96 Jan  4 08:48 .

~/Projects/test-rb  v3.4.2
$ export BUNDLE_FROZEN=true

~/Projects/test-rb  v3.4.2
$ cat Gemfile
gem "sinatra"

~/Projects/test-rb  v3.4.2
$ bundle install
The frozen setting requires a lockfile. Please make sure you have checked your
Gemfile.lock into version control before deploying.
```

> Looking at the added test it's for a second bundle install done later, but in that case it's much simpler to just bundle config set frozen true yourself after setup-ruby.

Makes sense.

Might an option like this make sense for the case where you set `bundler-cache: true` since it runs `bundle install`? I suppose the workaround would be adding an `env` line with `BUNDLE_FROZEN=true` globally in the workflow or in the specific step, e.g.,

```yaml
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: '3.4'
        env:
          BUNDLE_FROZEN: true
        bundler-cache: true
```

PR #836, opened 2025-12-13
When setting $JAVA_HOME also add $JAVA_HOME/bin to PATH

eregon:
* See discussion in https://github.com/rake-compiler/rake-compiler/pull/247
  it is messy if that's not the case and also creates inconsistencies based on which Java tool/executable is used.


eregon:
It works well, and Node.js `path.join` does the right thing:
https://github.com/ruby/setup-ruby/actions/runs/20197116199/job/57982688126?pr=836#step:3:36
```
Adding C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\21.0.9-10.0\x64\bin to Path
...
Run # Show PATH with Powershell
C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\21.0.9-10.0\x64\bin
D:\jruby-10.0.2.0\bin
```

PR #837, opened 2025-12-17
Add ruby-3.4.8

stephanmoneybird:


eregon:
Thank you but I'll use the automatic PR, it wasn't created yet because macos-13 is gone, fixing that: https://github.com/ruby/ruby-builder/actions/runs/20294518384/job/58285316067
https://github.com/ruby/setup-ruby/pull/838

PR #851, opened 2026-01-06
Add token input for downloading release assets

TingluoHuang:
This pull request adds support for authenticating downloads of Ruby release assets from GitHub using a token, helping to avoid rate limiting and enabling compatibility with GitHub Enterprise Server (GHES). The changes introduce a new `token` input, update documentation, and ensure the token is used when downloading Ruby and related build tools.

Fixes https://github.com/ruby/setup-ruby/issues/852

### DIFF
--- a/common.js
+++ b/common.js
@@ -442,3 +442,8 @@ export function isExactCacheKeyMatch(key, cacheKey) {
       }) === 0
   );
 }
+
+export async function download(url) {
+  const auth = inputs.token ? `token ${inputs.token}` : undefined
+  return await tc.downloadTool(url, undefined, auth)
+}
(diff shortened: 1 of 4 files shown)


eregon:
Could the default token always be good enough?
Have you seen a failure with GHES? Would no token be good enough for GHES?

TingluoHuang:
The default token should be good for github.com, we have similar usage for [setup-node](https://github.com/actions/setup-node/blob/main/action.yml#L18)
For GHES, customer can leave it empty so they are doing un-authed just like today, or they can provide a github.com PAT to match experience with running this Action in github.com

PR #854, opened 2026-01-07
Generate test matrix dynamically

ntkme:
This PR moves from a hard coded test matrix to a generated test matrix, to reduce the maintenance overhead when new runners or releases are introduced.


eregon:
Thank you a lot for this!
I checked the list of generated jobs and it seems match exactly the old jobs.

PR #856, opened 2026-01-10
Add restriction and validation for download urls

ntkme:
This PR takes a slightly different approach to the URL safety guardrail discussed in #848:

1. It adds the check into directly `generate-windows-version.rb` in the first place so hopefully we can make sure the bot generated PR are clean. - The possibility of tempering `generate-windows-version.rb` itself shall be guard via checking files that has been updated by bot PR. 
2. The `generate-windows-version.rb` added to run as part of the normal CI as well, and the output is compared with checked-in version, to make sure avoid the case where bot is compromised and the windows versions files are tempered without modifying `generate-windows-version.rb` (of course, hacker first need gain access to other repos to upload compromised artifacts).

This might introduce a little bit noise if a new RI2 release has been released but has not been added here yet, and around the same time someone tries to contribute other changes. I think that's going to be extremely rare, and it's not a huge issue anyways.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -370,6 +370,7 @@ jobs:
         run: ruby -e "abort if File.exist?('C:/msys64/mingw64/bin/gcc.exe')"
 
   validate-windows-versions:
+    name: "Check windows*versions.json are up-to-date"
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


eregon:
Nice idea.
I guess the main downside is this new CI job would fail when there is a new RubyInstaller release but it's not merged yet, e.g. for other PRs that run CI around that time. Might not be a big deal though, and we can merge manually anyway if it gets in the way.

eregon:
I like this idea, let's try it.

tcka4enckoalexei2014-commits:
1

PR #861, opened 2026-01-18
Fix compatibility to ruby-3.2

larskanis:
... which is the default ruby version on Github Actions. `URI::Generic#to_str` was added in ruby-3.3.

That made the windows new version check fails since 2026-01-13.

Also add the new RubyInstaller releases 4.0.1-1 and 3.2.10-1 to make the PR succeed.


eregon (approved):
Thanks!

eregon:
> which is the default ruby version on Github Actions

The check in this repo CI uses `ruby-version: ruby` but I see, you mean https://github.com/ruby/ruby-builder/actions/workflows/check-new-windows-versions.yml
We should fix that to use latest release too for consistency.
But actually I'm thinking to move that CI job here, I don't think it makes sense to have it in that repo anymore.

PR #865, opened 2026-01-23
add hook for setup script

binford2k:
This gives you the ability to pass script text to the `setup-ruby` action. The script will be executed after Ruby and Bundler are installed, but before `bundle install` is run.

This lets you install packages or Bundler plugins, download patches, anything.

If you want to execute a script on disk, just pass the filename, like `'./setup.sh'` or `'ruby setup.rb'`

If you intend to use this to install a custom Bundler version or the like, make sure to specify `bundler: none` so you don't get a second version installed.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -15,6 +15,11 @@ inputs:
       For 'latest', `gem update --system` is run to update to the latest compatible RubyGems version.
       Ruby head/master builds will not be updated.
       Similarly, if a version number is given, `gem update --system <version>` is run to update to that version of RubyGems, as long as that version is newer than the one provided by default.
+  setup-script:
+    description: |
+      Setup code to be executed after Ruby and Bundler are installed, but before `bundle install` is run.
+      If you want to execute a script on disk, just pass the filename, like `'./setup.sh'` or `'ruby setup.rb'`
+      If you intend to use this to install a custom Bundler version or the like, make sure to specify `bundler: none`.
   bundler:
     description: |
       The version of Bundler to install. Either 'Gemfile.lock' (the default), 'default', 'latest', 'none', or a version number (e.g., 1, 2, 2.1, 2.1.4).
--- a/dist/index.js
+++ b/dist/index.js
@@ -92443,6 +92443,7 @@ const inputDefaults = {
   'rubygems': 'default',
   'bundler': 'Gemfile.lock',
   'bundler-cache': 'false',
+  'setup-script': '',
   'working-directory': '.',
   'cache-version': bundler.DEFAULT_CACHE_VERSION,
   'self-hosted': 'false',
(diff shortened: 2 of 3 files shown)


eregon:
See https://github.com/ruby/setup-ruby/pull/239#issuecomment-987785246

binford2k:
The arguments in that thread don't really hold water, imho.

If you want to run a specific shell, then you can do so with e.g. `bash -c '...'`. If you want to do strange things like pipes and redirects and whatnot, then you put them in a shell script, as already discussed. Declare the convention that this will fail fast and the first command that fails returns non-zero and then there's no ambiguity.

I agree that the marginally better solution would be to separate this into two actions, but that's a lot of work to do and a lot of work to migrate users to.

There are use cases for injecting commands prior to `bundle install`, evidenced by the fact that it keeps coming up. My own immediate use case is that I need to run `bundle plugin install` commands* prior to `bundle install` and there's no good way to do that without giving up most of the benefit of using this action to begin with. Moreover, if Bundler plugins are improved and become more common in the ecosystem like the Bundler team seems to have indicated, then this kind of use case will just become more common.

There are plenty of other use cases though, like setting up a build/install environment more complex than `setup-ruby-pkgs` can handle or such. Adding a simple hook here multiplies the usefulness of this awesome action with only a very small increase in complexity.

> * no, I can't just add them to the `Gemfile`. That would defeat the purpose of being able to test Puppet modules unmodified. It would also be immensely more fragile to automate.

eregon:
> If you want to run a specific shell, then you can do so with e.g. `bash -c '...'`. If you want to do strange things like pipes and redirects and whatnot, then you put them in a shell script, as already discussed. Declare the convention that this will fail fast and the first command that fails returns non-zero and then there's no ambiguity.

That will still misbehave for several cases, including shell heredocs, line continuations with `\`, `&&`, etc.

Regarding the shell, maybe some people want the same as the default shell per platform, and some want Bash on all platforms. And it seems clear nobody wants to write `bash -c '...'` or the powershell equivalent, it would quickly get out of hand regarding quoting.

Basically it feels like a feature that always asks for more edge cases and infinite work to keep improving it.

> If you intend to use this to install a custom Bundler version or the like, make sure to specify `bundler: none` so you don't get a second version installed.

I don't think this would work, `bundler-cache:` needs to know the Bundler version.

For your use-case would this work? Can you try it?
```yaml
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: '3.4'
    - run: custom commands
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: '3.4'
        bundler-cache: true
```

binford2k:
No, it doesn't work. I need to configure something _before_ the `bundle install`

eregon:
`- run: custom commands` above runs before the `bundle install`.

PR #897, opened 2026-03-27
Refactor matrix script

ntkme:
@eregon Please take a look if this looks good to you.

### DIFF
--- a/generate-test-matrix.rb
+++ b/generate-test-matrix.rb
@@ -33,9 +33,8 @@ runners = %w[
 macos_runners = runners.select { |runner| runner.start_with?('macos-') }
 ubuntu_runners = runners.select { |runner| runner.start_with?('ubuntu-') }
 windows_runners, non_windows_runners = runners.partition { |runner| runner.start_with?('windows-') }
-macos_intel_runners = runners.grep(/macos.+intel/)
 
-macos_arm64_runners, _macos_x64_runners = macos_runners.partition { |runner| !runner.end_with?('-intel')}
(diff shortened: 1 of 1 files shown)


eregon (approved):
Thank you!
I totally missed that variable 😅

eregon:
BTW do you have an idea how to apply this to all truffleruby/truffleruby+graalvm versions >= 34 (so `head`, `34`, `40`, etc), although `34`, `40` are not out yet, but it'd be nice if CI doesn't break when they are.

eregon:
Ah given `truffleruby_versions = %w[truffleruby truffleruby-head truffleruby+graalvm truffleruby+graalvm-head]` it's actually trivial, we just need to wait for 34 to be out then, and it'll be fixed for versions after that.

(sorry, it's late here)

ntkme:
In theory, it can be done by checking the latest version available in `ruby-builder-versions.json`, and generating the matrix differently. However, I think trying to handle a future change before it comes out seems to be a bit over-engineered for a simple one-time issue. 

---

As 34 come out, as part of the ruby-builder-bot PR, we can just manually add a commit to have it changed from:
``` ruby
truffleruby_versions = %w[truffleruby truffleruby-head truffleruby+graalvm truffleruby+graalvm-head]
matrix += non_windows_runners.product(truffleruby_versions)
```
To:
``` ruby
truffleruby_versions = %w[truffleruby truffleruby-head truffleruby+graalvm truffleruby+graalvm-head]
matrix += (macos_arm64_runners + ubuntu_runners).product(truffleruby_versions)
```

eregon:
Nice, done in https://github.com/ruby/setup-ruby/pull/901

PR #905, opened 2026-04-21
Allow customising project ID in cache key rather than it always including working directory

### LINKED ISSUE #904: Bundler cache not working due to ephemeral workdir

- [x] I verified it reproduces with the latest version with `- uses: ruby/setup-ruby@v1` (see [Versioning policy](https://github.com/ruby/setup-ruby/blob/master/README.md#versioning))

- [x] ~I tried to reproduce the issue locally by following the workflow steps (including all commands done by `ruby/setup-ruby`, except for `Downloading Ruby` & `Extracting Ruby`),
and it did not reproduce locally (if it does reproduce locally, it's not a ruby/setup-ruby issue)~

self-hosted runner

N/A

### Any other notes?

We use GitHub Actions runners on AWS CodeBuild with a custom Docker image. The Bundler gems caching is unfortunately not working for us - Bundler installs all gems from scratch on every run. I had a look into why, and noticed a difference in the cache key between jobs, e.g.:

```
# Run 1
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.3.6-wd-/codebuild/output/src3491982091/src/actions-runner/_work/better_caring/better_caring-with--without--only--Gemfile.lock-3f96ad38961e7a44e242400b976a8833512d7d013bd35c5f24b09672d2fd2667
# Run 2
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.3.6-wd-/codebuild/output/src1813367680/src/actions-runner/_work/better_caring/better_caring-with--without--only--Gemfile.lock-3f96ad38961e7a44e242400b976a8833512d7d013bd35c5f24b09672d2fd2667
```

That unfortunately results in being unable to reuse a previous cache.

The cause of the cache key difference is the differing working directory. This looks to be done by CodeBuild for run isolation, which seems fair. I vaguely recall seeing a similar thing with Buildkite.

I looked into the source code of this action to see if the cache key could be configured, and I don't see a way to do that. It seems there is no way to avoid the inclusion of the ephemeral working directory path in the cache key:

https://github.com/ruby/setup-ruby/blob/7372622e62b60b3cb750dcd2b9e32c247ffec26a/bundler.js#L190-L196

One option for us could be to change to using a constant workdir; but I'm not confident we won't run into run isolation problems from doing that.

Please consider allowing omission of the workdir from the cache key. I'd suggest replacing it with another way to identify the project - perhaps the Git remote URL (with special characters substituted as necessary).

Thank you for your work on this project =)

ZimbiX:
Resolves https://github.com/ruby/setup-ruby/issues/904

I decided to go with something simple rather than using the Git repo remote URL which I'd first thought to use. The URL might not suit all use-cases - perhaps someone has multiple bundles in the one repo (hence you having added the workdir to distinguish). Instead, a simple customisable string value `project-id` would allow using the action multiple times in a job for different directories with a different `project-id`.

## Testing

Tested and working in our GitHub Actions workflow using:

```yaml
      # Fork of ruby/setup-ruby to support ephemeral workdir
      - uses: ZimbiX/setup-ruby@remove-pwd-from-bundler-cache-key
        with:
          ruby-version: 3.3.6
          bundler-cache: true
          project-id: better_caring
```

Logs:

```
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.3.6-better_caring-with--without--only--Gemfile.lock-3f96ad38961e7a44e242400b976a8833512d7d013bd35c5f24b09672d2fd2667
Cache hit for: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.3.6-better_caring-with--without--only--Gemfile.lock-3f96ad38961e7a44e242400b976a8833512d7d013bd35c5f24b09672d2fd2667
Received 0 of 140644053 (0.0%), 0.0 MBs/sec
Received 0 of 140644053 (0.0%), 0.0 MBs/sec
Received 8388608 of 140644053 (6.0%), 2.7 MBs/sec
Received 46137344 of 140644053 (32.8%), 11.0 MBs/sec
Received 100663296 of 140644053 (71.6%), 19.2 MBs/sec
Received 136449749 of 140644053 (97.0%), 21.7 MBs/sec
Received 140644053 of 140644053 (100.0%), 20.8 MBs/sec
Cache Size: ~134 MB (140644053 B)
/usr/bin/tar -xf /codebuild/output/src2839699089/src/actions-runner/_work/_temp/e8390f69-4590-46ba-aa6f-789b62f5e608/cache.tzst -P -C /codebuild/output/src2839699089/src/actions-runner/_work/better_caring/better_caring --use-compress-program unzstd
Cache restored successfully
Found cache for key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.3.6-better_caring-with--without--only--Gemfile.lock-3f96ad38961e7a44e242400b976a8833512d7d013bd35c5f24b09672d2fd2667
```

### DIFF
--- a/bundler.js
+++ b/bundler.js
@@ -155,7 +155,7 @@ function bundlerConfigSetArgs(bundlerVersion, key, value) {
   }
 }
 
-export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVersion, bundlerVersion, cacheVersion) {
+export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVersion, bundlerVersion, cacheVersion, projectId) {
   if (gemfile === null) {
     console.log('Could not determine gemfile path, skipping "bundle install" and caching')
     return false
@@ -189,7 +189,7 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
 
   // cache key
   const paths = [cachePath]
-  const baseKey = await computeBaseKey(engine, rubyVersion, lockFile, cacheVersion)
+  const baseKey = await computeBaseKey(engine, rubyVersion, lockFile, cacheVersion, projectId)
   const key = `${baseKey}-${await common.hashFile(lockFile)}`
   // If only Gemfile.lock changes we can reuse part of the cache, and clean old gem versions below
   const restoreKeys = [`${baseKey}-`]
@@ -242,12 +242,12 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
   return true
 }
 
-async function computeBaseKey(engine, version, lockFile, cacheVersion) {
-  const cwd = process.cwd()
+async function computeBaseKey(engine, version, lockFile, cacheVersion, specifiedProjectId) {
+  const projectId = specifiedProjectId || `wd-${process.cwd()}`
   const bundleWith = process.env['BUNDLE_WITH'] || ''
   const bundleWithout = process.env['BUNDLE_WITHOUT'] || ''
   const bundleOnly = process.env['BUNDLE_ONLY'] || ''
-  let key = `setup-ruby-bundler-cache-v6-${common.getOSNameVersionArch()}-${engine}-${version}-wd-${cwd}-with-${bundleWith}-without-${bundleWithout}-only-${bundleOnly}`
+  let key = `setup-ruby-bundler-cache-v6-${common.getOSNameVersionArch()}-${engine}-${version}-${projectId}-with-${bundleWith}-without-${bundleWithout}-only-${bundleOnly}`
 
   if (cacheVersion !== DEFAULT_CACHE_VERSION) {
     key += `-v-${cacheVersion}`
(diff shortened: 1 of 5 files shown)


bundler.js
```diff
@@ -242,12 +242,12 @@ export async function bundleInstall(gemfile, lockFile, platform, engine, rubyVer
   return true
 }
 
-async function computeBaseKey(engine, version, lockFile, cacheVersion) {
-  const cwd = process.cwd()
+async function computeBaseKey(engine, version, lockFile, cacheVersion, specifiedProjectId) {
+  const projectId = specifiedProjectId || `wd-${process.cwd()}`
```
eregon, line 246: This would ignore the user-supplied `working-directory` input: https://github.com/ruby/setup-ruby/blob/master/action.yml

More generally I think we don't need a project ID here, I'll take a look at #907 next.

ZimbiX:
Solution: https://github.com/ruby/setup-ruby/issues/904#issuecomment-5550952696

Issue #915, opened 2026-05-20
ruby-builder build failed with 4.0.5


GeoffTidey:
https://github.com/ruby/ruby-builder/actions/runs/26135567982/job/76869926439

(needs a rerun?)

batter:
+1, the build succeeded but the push of the build to the cloud failed and it's preventing the ability to utilize Ruby 4.0.5 on GitHub Actions

ntkme:
@eregon @hsbt The failure looks like a temporary network glitch. Can you please trigger an action re-run when you get a chance? Most of the maintainers for setup-ruby do not have access to re-run failed ruby-builder jobs.

eregon:
Rerun: https://github.com/ruby/ruby-builder/actions/runs/26135567982/job/77043601975

eregon:
> the build succeeded but the push of the build to the ubuntu repo failed

That's not even close to the reality, is that AI-generated text?

ntkme:
4.0.5 has been released.

batter:
> > the build succeeded but the push of the build to the ubuntu repo failed
> 
> That's not even close to the reality, is that AI-generated text?

No, that was generated by my brain 😅.. I meant the CI action to push the build up to the cloud failed. I looked again and saw it failed due to it failing to pull dependencies from the ubuntu apt-get repo.

Thank you for resolving!

eregon:
OK no worries, I appreciate the reply.

PR #916, opened 2026-05-20
Add ruby 4.0.5

just1602:
I inspired myself from 6aaa311d81eba98ae12eaffbcb63296ace0efcde for this one, I hope it's fine.


ntkme:
Please be patient that this won't work before #915 is addressed. Also this PR is missing some necessary update.

I recommend wait for #915 to be fixed and our bot-user would automatically create a proper PR.

eregon:
Yes let's just use the automated PR since fixing this one would need extra work

PR #919, opened 2026-06-05
Use BUNDLE_LOCKFILE when detecting the lockfile

Thomascountz:
If a workflow sets `BUNDLE_LOCKFILE`, setup-ruby still assumes `${BUNDLE_GEMFILE}.lock` or `gems.locked`. That means it can read the wrong `BUNDLED WITH` version, fail to enable deployment mode because it checks the wrong lockfile path, and/or generate a cache key from the wrong lockfile.

Bundler added `BUNDLE_LOCKFILE` support in [PR #9059](https://github.com/ruby/rubygems/pull/9059)

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ should be able to fix them by setting `rubygems: 3.0.0` or higher.
 
 By default, Bundler is installed as follows:
 
-* If there is a `Gemfile.lock` file (or `$BUNDLE_GEMFILE.lock` or `gems.locked`) with a `BUNDLED WITH` section,
+* If there is a `Gemfile.lock` file (or `$BUNDLE_LOCKFILE`, `$BUNDLE_GEMFILE.lock`, or `gems.locked`) with a `BUNDLED WITH` section,
   that version of Bundler will be installed and used.
 * If the Ruby ships with Bundler 2.2+ (as a default gem), that version is used.
 * Otherwise, the latest compatible Bundler version is installed (Bundler 2 on Ruby >= 2.3, Bundler 1 on Ruby < 2.3).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -168,7 +168,7 @@ should be able to fix them by setting `rubygems: 3.0.0` or higher.
 
 By default, Bundler is installed as follows:
 
-* If there is a `Gemfile.lock` file (or `$BUNDLE_GEMFILE.lock` or `gems.locked`) with a `BUNDLED WITH` section,
+* If there is a `Gemfile.lock` file (or `$BUNDLE_GEMFILE.lock`, `$BUNDLE_LOCKFILE`, or `gems.locked`) with a `BUNDLED WITH` section,
```
eregon: In terms of ordering in the code `BUNDLE_LOCKFILE` has precedence so let's do that here too and below as well
```suggestion
* If there is a `Gemfile.lock` file (or `$BUNDLE_LOCKFILE`, `$BUNDLE_GEMFILE.lock`, or `gems.locked`) with a `BUNDLED WITH` section,
```

eregon (commented):
Looks good, thank you

bundler.js
```diff
@@ -15,14 +15,16 @@ function isValidBundlerVersion(bundlerVersion) {
 // The returned gemfile is guaranteed to exist, the lockfile might not exist
 export function detectGemfiles() {
   const gemfilePath = process.env['BUNDLE_GEMFILE'] || 'Gemfile'
+  const lockfilePath = process.env['BUNDLE_LOCKFILE']
+
   if (fs.existsSync(gemfilePath)) {
-    return [gemfilePath, `${gemfilePath}.lock`]
+    return [gemfilePath, lockfilePath || `${gemfilePath}.lock`]
```
ntkme, line 21: Given that we have a `fs.existsSync` check for `BUNDLE_GEMFILE` or its default value `Gemfile` that would thrown an exception if missing, we should probably have `fs.existsSync` check for `BUNDLE_LOCKFILE` if it's explicitly set. In this case, unlike the check for `gemfilePath`, we should not check existence of `Gemfile.lock` because it's optional.

eregon: I'm not following, the code is fine as-is, no? Because the lockfile is always optional.

eregon (approved):

bundler.js
ntkme: If you explicit define an environment variable for it, most likely it means you want to explicitly use that file. In that case if the file specified in the variable does not exist and we treat it as optional, wouldn’t that be confusing?

eregon: Since it's a variable recognized by Bundler people might use it independently of setup-ruby so I think it's fine and not something we should check.

But, that makes me think probably this needs to be part of the cache key.

eregon: lockFile is already in computeBaseKey, so all good

PR #923, opened 2026-06-20
Add support for ubuntu-26.04 and ubuntu-26.04-arm

### LINKED ISSUE #922: Add Ubuntu 26.04 support
GitHub recently launched Ubuntu 26.04 support in public preview.

https://github.com/actions/runner-images/issues/14226

Currently it fails with following:

> Error: The current runner (ubuntu-26.04-x64) was detected as self-hosted because the platform does not match a GitHub-hosted runner image (or that image is deprecated and no longer supported).

eregon:
Fixes https://github.com/ruby/setup-ruby/issues/922

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ The action works on these [GitHub-hosted runners](https://docs.github.com/en/act
 
 | Operating System | Supported |
 | ---------------- | --------- |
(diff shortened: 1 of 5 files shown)


eregon:
Merging, for the `head` builds I need to merge this first to fix the cycle as ruby-dev-builder uses setup-ruby for BASERUBY.

PR #934, opened 2026-09-03
Add Debian 12 and 13 support

brodock:
Download the matching Ubuntu ruby-builder binaries instead of treating Debian as an unsupported self-hosted image, and install missing CRuby runtime libraries with apt-get when the job is root.


eregon:
The rule in this action is: if it's a platform we can't test easily and reliably (i.e. not a GitHub-hosted runner), then it's not supported and might only work as a self-hosted runner.
Installing system packages at runtime is also explicitly against the design of this action.
I also wouldn't want to confuse users by talking about Debian in the README.

